PR #2, opened 2017-10-23
Add elastic client signer

thibautcornolti:
Add function allowing to sign each request for AWS ElasticSearch.


es/clientsigner.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"errors"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/signer/v4"
+	"github.com/sha1sum/aws_signing_client"
+	"gopkg.in/olivere/elastic.v5"
+	"strings"
+)
+
+func checkError(splittedEndpoint []string, creds *credentials.Credentials) error {
```
schubev: “splitted” doesn’t exist in english. The past participle for “to split” is “split”.

schubev: What does “checkError” check errors for ? This is a very generic function name.

es/clientsigner.go
```diff
@@ -0,0 +1,62 @@
...
+	if creds == nil {
+		return errors.New("Nil credential")
+	} else if _, err := creds.Get(); err != nil {
+		return err
+	} else if len(splittedEndpoint) <= 4 {
```
schubev: What is the split endpoint ? I figure it is an endpoint that should have at least five parts ( which isn’t made obvious by using the number 4 ), but I don’t know what it is used for.

es/clientsigner.go
```diff
@@ -0,0 +1,62 @@
...
+	"errors"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/signer/v4"
+	"github.com/sha1sum/aws_signing_client"
+	"gopkg.in/olivere/elastic.v5"
+	"strings"
+)
+
+func checkError(splittedEndpoint []string, creds *credentials.Credentials) error {
+	if creds == nil {
```
schubev: Do credentials still work without being a pointer ? Are they small in memory ? If both statements are true, perhaps you should not be using a pointer, it complicate the code with this sort of checks. Or you can assume it is the caller’s responsibility to provide a non-nil pointer.

es/clientsigner.go
schubev: Please document the function, let’s have all but the most trivial functions documented.

es/clientsigner.go
schubev: Also, why 4 ? Consider making this a `const` with a GoDoc.

es/clientsigner.go
```diff
@@ -0,0 +1,62 @@
...
+		return errors.New("Wrong endpoint parameter")
+	}
+	return nil
+}
+
+// NewSignedElasticClient create a signed *elastic.Client ready for a using with AWS ElasticSearch.
+// It takes as parameter:
+//		- endpoint: The endpoint URI gettable from AWS.
+//		- credentials: Credentials from AWS/Credentials.
+func NewSignedElasticClient(endpoint string, credentials *credentials.Credentials) (*elastic.Client, error) {
```
schubev: Good function, it does only one thing, and its purpose is clear. You should just reverse the order of the return values, the Go convention is to put the error last.

schubev: Does `elastic.Client` still work if passed as a value ? In that case it would make for simpler code if you dropped the pointer and used the value directly, unless it is a large structure. As long as you guarantee that the returned client is valid iff the error is nil, there is no problem returning a value with an error.

es/clientsigner.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, err
+	}
+	prefix := ""
+	if !strings.HasPrefix(endpoint, "http") {
+		prefix = "https://"
+	}
+	return elastic.NewClient(
+		elastic.SetURL(prefix+endpoint),
+		elastic.SetScheme("https"),
+		elastic.SetHttpClient(awsClient),
```
schubev: So you can plug an HTTP client into an ElasticSearch client, that’s good !

es/clientsigner.go
```diff
@@ -0,0 +1,62 @@
...
+		elastic.SetSniff(false),
```
schubev: What does `SetSniff` do ? Does it default as `true` ?

es/clientsigner.go
```diff
@@ -0,0 +1,62 @@
...
+		return errors.New("Nil credential")
+	} else if _, err := creds.Get(); err != nil {
+		return err
+	} else if len(splittedEndpoint) <= 4 {
+		return errors.New("Wrong endpoint parameter")
+	}
+	return nil
+}
+
+// NewSignedElasticClient create a signed *elastic.Client ready for a using with AWS ElasticSearch.
```
giubil: `for a using with AWS ElasticSearch` is not English

es/clientsigner.go
giubil: Please write unit tests for this function

giubil: `NewSignedElasticClient` does not contain a verb. `GetNewSignedElasticClient` would be a better suited function name

giubil (changes requested):

es/clientsigner.go
```diff
@@ -0,0 +1,62 @@
...
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+package es
+
+import (
```
schubev: Please order your imports in three parts, separated by empty lines : standard library imports, non-standard library imports, and finally, imports from the current repository.

schubev (changes requested):

es/clientsigner.go
schubev: `New*` functions seem to be idiomatic of Go. I think it is understandable enough.

giubil: Okay my bad

es/clientsigner.go
giubil: `SetSniff` will set if sniffing is enabled or not. Sniffing allows for auto-discovery of ElasticSearch nodes in a cluster.
It is disabled because of this note in the elastic go library : https://github.com/olivere/elastic/wiki/Using-with-AWS-Elasticsearch-Service#sniffing

es/clientsigner.go
thibautcornolti: According to http://docs.aws.amazon.com/sdk-for-go/api/aws/credentials, all functions allowing to get a credentials return a pointer. That's why I use pointers.

schubev:
Please rename _clientsigner.go_ to _awsclientsigner.go_ and I believe that PR is ready to go.

schubev (approved):

giubil (approved):

PR #3, opened 2017-10-23
Add human parser for es output

thibautcornolti:
Add parser to transform a row SearchResult to a refine map[string]interface{} more human readable.


costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+		flatSubAggregation(&bucketAggregation)
+		result[name] = bucketAggregation
+	} else {
+		result[aggregationName] = root
+	}
+}
+
+// GetParsedElasticSearchResult is used to parse an *elastic.SearchResult
+// to a human readable map[string]interface{} that will be able to be
+// transform to JSON and send to the UI
```
giubil: `to be transform` is not english

giubil: Please write unit tests for `GetParsedElasticSearchResult`

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"gopkg.in/olivere/elastic.v5"
+)
+
+func flatSubAggregation(in *interface{}) {
```
giubil: The function name `flatSubAggregation` does not seem English. `flattenSubAggregation` would be better

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+			flatSubAggregation(&aggregationType)
+		}
+		in[skey] = aggregationType
+		delete(in, "key")
+		delete(in, "key_as_string")
+	}
+	delete(in, aggregationName)
+}
+
+func recurMap(in map[string]interface{}) {
```
giubil: This function name is not explicit enough

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+		if aggregationName[0] == '&' {
+			recurTab(&aggregationType)
+		} else if aggregationName[0] == '*' {
+			recurMap(field)
+		}
+		cleanSubAggregation(aggregationName, aggregationType, in)
+	}
+}
+
+func recurTab(in *interface{}) {
```
giubil: This function name is not explicit enough

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+		delete(field, "doc_count")
+		recurMap(field)
+	}
+	flatSubAggregation(in)
+}
+
+func prepareRecursiveParsing(result map[string]interface{}, aggregationName string, aggregation *json.RawMessage) {
+	var root map[string]interface{}
+	err := json.Unmarshal(*aggregation, &root)
+	if err != nil {
```
giubil: An error is created, but the `jsonlog` logger is not used, and the function cannot return an error. Please add logging and error returning

giubil (changes requested):

costs/parser.go
schubev: Please add Godoc to your functions and ensure the names are descriptive. This function is not trivial and needs this.

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+	newMap := make(map[string]interface{})
+	for _, field := range oldTab {
+		for name, subField := range field.(map[string]interface{}) {
+			newMap[name] = subField
+		}
+	}
+	*in = newMap
+}
+
+func cleanSubAggregation(aggregationName string, aggregationType interface{}, in map[string]interface{}) {
```
schubev: What do you mean by “clean” ? What is the `in` parameter ?

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+	key, ok := in["key_as_string"]
+	if !ok {
+		key, ok = in["key"]
+	}
+	skey := fmt.Sprintf("%v", key)
```
schubev: You only use `skey` in your second `if` block. Why not initialize it there, and transform your `if !ok {} ... if ok {}` by an `if !ok {} else {}` ?

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+	if ok {
+		if aggregationMapType, ok := aggregationType.(map[string]interface{}); ok && aggregationName[0] == '*' {
```
schubev: Consider using abbreviations for longer names, as long as it doesn’t obscure them. `aggr` for `aggregation` is fine.

costs/parser.go
schubev: I agree.

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	delete(in, aggregationName)
+}
+
+func recurMap(in map[string]interface{}) {
+	for aggregationName, field := range in {
+		if aggregationName[0] != '&' && aggregationName[0] != '*' {
+			continue
+		}
+		field := field.(map[string]interface{})
```
schubev: Is it guaranteed that `field` is a `map[string]interface{}` ?

costs/parser.go
schubev: I agree.

costs/parser.go
schubev: Yes. The best thing would be that your function take in a `context.Context`, and use `jsonlog.LoggerFromContextOrDefault(ctx context.Context) jsonlog.Logger` to get the logger.

costs/parser.go
```diff
@@ -0,0 +1,114 @@
...
+		result[name] = bucketAggregation
+	} else {
+		result[aggregationName] = root
+	}
+}
+
+// GetParsedElasticSearchResult is used to parse an *elastic.SearchResult
+// to a human readable map[string]interface{} that will be able to be
+// transform to JSON and send to the UI
+func GetParsedElasticSearchResult(esResult *elastic.SearchResult) map[string]interface{} {
```
schubev: What do you mean by Parsed ? Serializable to JSON ? If so, I would suggest naming it `GetJsonSimplifiedElasticSearchResult`.

schubev (changes requested):

costs/parser.go
```diff
@@ -51,7 +51,7 @@ func cleanSubAggregation(aggregationName string, aggregationType interface{}, in
 	delete(in, aggregationName)
 }
 
-func recurMap(in map[string]interface{}) {
+func browseRecursivlyMap(in map[string]interface{}) {
```
giubil: `browseMapRecursivly` sound more correct to me than `browseRecursivlyMap`

costs/parser.go
```diff
@@ -61,54 +61,57 @@ func recurMap(in map[string]interface{}) {
...
-			recurMap(field)
+			browseRecursivlyMap(field)
 		}
-		cleanSubAggregation(aggregationName, aggregationType, in)
+		stripAggregationArchitecture(aggregationName, aggregationType, in)
 	}
 }
 
-func recurTab(in *interface{}) {
+func browseRecursivlyTab(in *interface{}) {
```
giubil: `browseTabRecursivly` sounds more correct to me than `browseRecursivlyTab`

costs/parser.go
thibautcornolti: It's not the same `ok`, if `ok` is `false`, `ok` will change and I have to recheck it.

costs/parser.go
thibautcornolti: Absolutely since the user will tag the name of its aggregations with `*` for a map and `&` for a tab.

costs/parser.go
```diff
@@ -0,0 +1,118 @@
...
+			flattenSubAggregation(&aggrType)
+		}
+		in[skey] = aggrType
+		delete(in, "key")
+		delete(in, "key_as_string")
+	}
+	delete(in, aggrName)
+}
+
+func browseMapRecursivly(in map[string]interface{}) {
```
schubev: `browseMapRecursively` and `browseTabRecursively` have no return value, so I expect they have side effects. Please make sure whenever functions have side effects these are clearly documented, preferably in the function name. I have no idea what sort of side effects browsing a map can have.

schubev (changes requested):
Please add godoc to your functions.

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+			browseTabRecursivly(&aggrType)
+		} else if aggrName[0] == '*' {
+			browseMapRecursivly(field)
+		}
+		stripAggregationArchitecture(aggrName, aggrType, in)
+	}
+}
+
+// browseTabRecursivly browses a map and calls browseMapRecursivly.
+// In the end it calls flattenSubAggregation to flat the array.
```
giubil, line 78: `to flatten`

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+		for name, subField := range field.(map[string]interface{}) {
+			newMap[name] = subField
+		}
+	}
+	*in = newMap
+}
+
+// stripAggregationArchitecture strips an array by
+// deleting some values and calling flattenSubAggregation.
+func stripAggregationArchitecture(aggrName string, aggrType interface{}, in map[string]interface{}) {
```
giubil, line 40: Ditto here for a function that does not return

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"gopkg.in/olivere/elastic.v5"
+)
+
+// flattenSubAggregation allows to transform an array
+// of map to one flattened map.
+func flattenSubAggregation(in *interface{}) {
```
giubil, line 27: This function does not return anything. Please either explicitly describe what this function modify, or return a value.
Functions that do not return and modify a parameter passed to it makes for quickly unmaintainable code.

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+		in[skey] = aggrType
+		delete(in, "key")
+		delete(in, "key_as_string")
+	}
+	delete(in, aggrName)
+}
+
+// browseMapRecursivly browses a map and calls browseMapRecursivly or
+// browseTabRecursivly, and stripAggregationArchitecture for each fields.
+func browseMapRecursivly(in map[string]interface{}) {
```
giubil, line 59: Ditto here for a function that does not return

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+		} else if aggrName[0] == '*' {
+			browseMapRecursivly(field)
+		}
+		stripAggregationArchitecture(aggrName, aggrType, in)
+	}
+}
+
+// browseTabRecursivly browses a map and calls browseMapRecursivly.
+// In the end it calls flattenSubAggregation to flat the array.
+func browseTabRecursivly(in *interface{}) {
```
giubil, line 79: Ditto here for a function that does not return

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+		field := field.(map[string]interface{})
+		delete(field, "doc_count")
+		browseMapRecursivly(field)
+	}
+	flattenSubAggregation(in)
+}
+
+// prepareRecursiveParsing prepares the first run of the recursive and
+// calls browseMapRecursivly or browseTabRecursivly.
+func prepareRecursiveParsing(result map[string]interface{}, aggrName string, aggr *json.RawMessage) error {
```
giubil, line 91: Ditto here for a function that does not return

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+		browseTabRecursivly(&aggrType)
+		result[name] = aggrType
+	} else if aggrName[0] == '*' {
+		browseMapRecursivly(root)
+		bucketAggr := aggrType.(map[string]interface{})["buckets"]
+		flattenSubAggregation(&bucketAggr)
+		result[name] = bucketAggr
+	} else {
+		result[aggrName] = root
+	}
```
giubil, line 111: This code here could be summed up to : 
```
/*
Magic happens
*/
```
Either use more explicit variable naming, comment inside the function, or redo the way the logic works.

giubil (changes requested):

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+	"strings"
+
+	"gopkg.in/olivere/elastic.v5"
+)
+
+// flattenSubAggregation allows to transform an array
+// of map to one flattened map.
+func flattenSubAggregation(in *interface{}) {
+	oldTab := (*in).([]interface{})
```
schubev, line 28: Can you guarantee that `*in` is an `[]interface{}` ? If it isn’t, please do some error handling with
```go
oldTab, ok := (*in).([]interface{})
```

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+	newMap := make(map[string]interface{})
+	for _, field := range oldTab {
+		for name, subField := range field.(map[string]interface{}) {
```
schubev, line 31: Why cast once more here, instead of directly casting `*in` to `[]map[string]interface{}` ? Also consider aliasing your types as I have no idea what such a type means.

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// stripAggregationArchitecture strips an array by
+// deleting some values and calling flattenSubAggregation.
+func stripAggregationArchitecture(aggrName string, aggrType interface{}, in map[string]interface{}) {
+	key, ok := in["key_as_string"]
+	if !ok {
+		key, ok = in["key"]
+	}
+	if skey := fmt.Sprintf("%v", key); ok {
```
schubev, line 45: Please reserve `if` initializations to cases where you initialize variables used in the condition. How about this ?
```go
key, ok := in["key_as_string"]
if !ok {
    key, ok = in["key"]
}
skey := fmt.Sprintf("%v", key) // are you sure about using "%v"?
if ok {
    …
}
```

schubev, line 45: With `key` being an `interface{}`, I have no idea what `"%v"` will result in. Please either give a type to your variables or create a function which explicitly handles the types `key` can have.

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+		delete(in, "key_as_string")
+	}
+	delete(in, aggrName)
+}
+
+// browseMapRecursivly browses a map and calls browseMapRecursivly or
+// browseTabRecursivly, and stripAggregationArchitecture for each fields.
+func browseMapRecursivly(in map[string]interface{}) {
+	for aggrName, field := range in {
+		if aggrName[0] != '&' && aggrName[0] != '*' {
```
schubev, line 61: What does it mean for an `aggrName` to begin with `&` or `*` ?

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+
+// GetJSONSimplifiedElasticSearchResult is used to parse an *elastic.SearchResult
+// to a human readable map[string]interface{} that will be able to be
+// transformed to JSON and send to the UI.
+// It will only parse aggregations with the name &buckets.{{name}} or *value.{{name}}
+// i.e. &buckets.regions and *value.cost.
+func GetJSONSimplifiedElasticSearchResult(esResult *elastic.SearchResult) (map[string]interface{}, error) {
+	res := make(map[string]interface{})
+	for name, aggregation := range esResult.Aggregations {
+		if err := prepareRecursiveParsing(res, name, aggregation); err != nil {
```
schubev, line 123: Why call the function `prepareRecursiveParsing` if it does the whole job ? When I see a `prepare*` function I expect that it should be followed with the logic which actually performs the job. Also this function seems to act on its arguments, which is usually Bad™. You can either change `prepareRecursiveParsing` to return its result, or define a type underlaid by `map[string]interface{}` and define methods over this type. At the very least, make sure it is _obvious_ that your function modifies its arguments when it does.

schubev (changes requested):

giubil:
Note : 
This code panic when the name of the bucket is `&{{name}}` instead of being `&bucket.{{name}}`

costs/es_request_parser.go
```diff
@@ -0,0 +1,128 @@
...
+		delete(in, "key")
+		delete(in, "key_as_string")
+	}
+	delete(in, aggrName)
+}
+
+// browseMapRecursivly browses a map and calls browseMapRecursivly or
+// browseTabRecursivly, and stripAggregationArchitecture for each fields.
+func browseMapRecursivly(in map[string]interface{}) {
+	for aggrName, field := range in {
```
giubil, line 60: This codes panics if the json key is empty.
This can happens if we aggregate on `regions` which can be empty for cost that is not region specific

schubev:
Rendered obsolete by 5a4c3b099eb0a1ba1fe25e915480486945ab6218.

PR #4, opened 2017-10-24
Integration of ElasticSearch request builder

giubil:
This integrates the ElasticSearch request builder, that will generate a `*elastic.SearchService` based on given parameters.

The code as is, is not merge ready. Indeed, there are no unit tests written for it.


costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+// A type switch is required for simulating downcasting from the interface elastic.Aggregation.
+// Current types on the type switch are TermsAggregation, FilterAggregation, SumAggregation and
+// DateHistogramAggregation.
+// If a new function creating a type that is not listed here is added to the paramNameToFuncPtr map
+// it should be added to the type switch, or the function will create bugged SubAggregations
+func nestAggregation(allAggrSlice []paramAggrAndName) elastic.Aggregation {
+	allAggrSlice = reverseAggregationArray(allAggrSlice)
+	aggrToNest := allAggrSlice[0]
+	for _, baseAggr := range allAggrSlice[1:] {
+		switch assertedBaseAggr := baseAggr.paramAggr.(type) {
```
thibautcornolti: You should use interfaces instead of a switch to cast.

thibautcornolti (changes requested):

costs/es_request_constructor.go
giubil: Because of the way the `SubAggregation` function is prototyped, it is not possible to add it to a new interface to be able to free ourselves from a type switch

costs/costs.go
```diff
@@ -22,8 +22,7 @@ import (
 
 // HandleRequest is a dummy request handler function. It does nothing except
 // some logging and returns static data.
-func HandleRequest(response http.ResponseWriter, request *http.Request) {
-	logger := jsonlog.LoggerFromContextOrDefault(request.Context())
+func HandleRequest(response http.ResponseWriter, request *http.Request, logger jsonlog.Logger) {
 	logger.Debug("Request headers.", request.Header)
```
schubev: Request handlers should have the following signature :
```go
func Handler(r *http.Request, a routes.Arguments) (int, interface{})
```
Also you should get the logger from the request context using `jsonlog.LoggerFromContextOrDefault`.

Since these are not part of the master repository yet, you should base this on the schubev/trackit repository instead, and rebase.

costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"gopkg.in/olivere/elastic.v5"
+)
+
+// paramNameToFuncPtr : map of string keys and functions pointer as values. For each possible param
```
schubev: This is not the correct GoDoc syntax. Your comment should start with _a full sentence_ whose first word is the identifier. Think that godoc will show just the first sentence when it does not have room for the full description, so it better be useful on its own and look like an actual sentence. For example :
```go
// paramNameToFuncPtr maps parameter names to the functions building the corresponding aggregation.
// …
```

costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+// after parsing (removing the ':<TAG_KEY>' in the case of the tag), there is a function associated to it
+// that create the Aggregations needed by this param.
+// If a new param, that is only creating aggregations, needs to be added,
+// a functions with the same prototype as the one below should be added to this map.
+var paramNameToFuncPtr = map[string]func([]string) []paramAggrAndName{
```
schubev: `func([]string) []paramAggrAndName` is not a descriptive type name, why not alias it ? It makes complex types more readable, for example `map[string]aggregationBuilder`.

costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+// defined by durationBegin and durationEnd
+func createQueryTimeRange(durationBegin time.Time, durationEnd time.Time) *elastic.RangeQuery {
+	return elastic.NewRangeQuery("usage_start_date").
+		From(durationBegin).To(durationEnd)
+}
+
+// createAggregationPerProduct : Creates and returns a new []paramAggrAndName of size 1 which creates a
+// bucket aggregation on the field 'product_name'
+func createAggregationPerProduct(_ []string) []paramAggrAndName {
+	res := make([]paramAggrAndName, 1)
```
schubev: Why not just this ?
```go
return []paramAggrAndName{
  paramAggrAndName{
    paramName: "product",
    paramAggr: elastic.NewTermsAggregation().Field("product_name").Size(aggregationMaxSize),
  }
}
```

costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+	"cost":    createCostSumAggregation,
+	"day":     createAggregationPerDay,
+	"week":    createAggregationPerWeek,
+	"month":   createAggregationPerMonth,
+	"year":    createAggregationPerYear,
+}
+
+// paramAggrAndName : Structure containing paramName and paramAggr, paramName being the name
+// of the parameter, such as "account". paramAggr is the Aggregation that correspond to that param
+// This struct is use to stoer all the aggregations and param name before nesting them.
```
schubev: `stoer`

costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+	"account": createAggregationPerAccount,
+	"tag":     createAggregationPerTag,
+	"cost":    createCostSumAggregation,
+	"day":     createAggregationPerDay,
+	"week":    createAggregationPerWeek,
+	"month":   createAggregationPerMonth,
+	"year":    createAggregationPerYear,
+}
+
+// paramAggrAndName : Structure containing paramName and paramAggr, paramName being the name
```
schubev: You talk about `param`, but I don’t know what it is a parameter for or what it means, or where it is used.

costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+// of the parameter, such as "account". paramAggr is the Aggregation that correspond to that param
+// This struct is use to stoer all the aggregations and param name before nesting them.
+type paramAggrAndName struct {
+	paramName string
```
schubev: So the type is named `paramAggrAndName`. I understand that whatever it contains applies to a `param`, whatever it is. You are repeating yourself by prepending `param` to the name of the member fields.

costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+	for i := len(aggregationArray)/2 - 1; i >= 0; i-- {
+		opp := len(aggregationArray) - 1 - i
+		aggregationArray[i], aggregationArray[opp] = aggregationArray[opp], aggregationArray[i]
+	}
+	return aggregationArray
+}
+
+// nestAggregation : takes a slice of paramAggrAndName type, and will nest the different aggregations.
+// Aggregations are nested by creating a chain of SubAggregation
+// A type switch is required for simulating downcasting from the interface elastic.Aggregation.
```
schubev: `to simulate downcasting`

costs/es_request_constructor.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+	}
+	return aggrToNest.paramAggr
+}
+
+// GetElasticSearchParams is used to construct an ElasticSearch *elastic.SearchService used to perform a request on ES
+// It takes as paramters :
+// 	- accountList []string : A slice of strings representing aws account number, in the format of the field
+//	'awsdetailedlineitem.linked_account_id'
+//	- durationBeing time.TIme : A time.Time struct representing the begining of the time range in the query
```
schubev: `TIme`

schubev (changes requested):
Looks good, please fix your documentations to be compliant with GoDoc. Also perhaps you could add a package-level documentation detailing the expected mapping for the ES index your requests would operate on.

costs/es_request_constructor.go
```diff
@@ -39,133 +45,143 @@ var paramNameToFuncPtr = map[string]func([]string) []paramAggrAndName{
...
 // 'tag.value'.
 // No SubAggregation is created in this fuction, as it needs to be created in the nestAggregation function
 func createAggregationPerTag(paramSplit []string) []paramAggrAndName {
 	res := make([]paramAggrAndName, 2)
 	res[0] = paramAggrAndName{
-		paramName: "tag_key",
-		paramAggr: elastic.NewFilterAggregation().
+		name: "tag_key",
+		aggr: elastic.NewFilterAggregation().
 			Filter(elastic.NewTermQuery("tag.key", fmt.Sprintf("user:%v", paramSplit[1])))}
```
thibautcornolti: Why don't you do exactly the same thing as above in your code ? I mean, you can return `[]paramAggrAndName{}` and put two `paramAggrAndName` inside.

thibautcornolti (changes requested):

costs/es_request_constructor.go
giubil: You're right, I didn't catch that, I will change it to be the same as the others `aggregationBuilder` functions

giubil:
This code is now ready to merge, as unit tests have been added.

Please note that right now the function `GetElasticSearchParams` (the only exported function) is not tested as a local ElasticSearch is needed to test it.

costs/es_request_constructor.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+}
+
+// createAggregationPerTag creates and returns a new []paramAggrAndName of size 2 which consits
+// of two aggregations that are required for the tag param.
+// The first aggregation is a FilterAggregation on the field 'tag.key', and with a value of
+// the tag key passed in the parameter 'paramSplit' in the form "user:<TAG_KEY_VALUE>".
+// The second aggregation is a TermsAggregation that creates bucket aggregation on the field
+// 'tag.value'.
+// No SubAggregation is created in this fuction, as it needs to be created in the nestAggregation function
```
thibautcornolti: You've written _fuction_

thibautcornolti (dismissed):

schubev (approved):

PR #9, opened 2017-11-07
Add pretty field in AWSAccount

thibautcornolti:
Add pretty field handling in the API.


giubil (dismissed):

schubev (approved):
Looks good to me. Though it reminds me that we now have to change the `External` to be `NOT NULL`.

giubil (approved):

PR #10, opened 2017-11-10
Adding handling of Variable URLs

thibautcornolti:
Allows to use Variable URLs as `/aws/:id` where :id is variable and can be getted in the handling final function.


routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
+package routes
```
schubev, line 1: You should comment your package.

routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
...
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+type (
+	parseIndex uint
+
+	ArgName string
+
```
schubev, line 13: You should comment your exported types and their exported members.

routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
...
+	ParseFloat32 struct {
+		ArgName ArgName
+	}
+	ParseFloat64 struct {
+		ArgName ArgName
+	}
+	ParseString struct {
+		ArgName ArgName
+	}
+	ParseVoid struct {
```
schubev, line 26: There is no such thing as `void` in Go, and you’re not parsing anything. Perhaps `Skip` would be more explicit ?

routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+type (
+	parseIndex uint
+
+	ArgName string
```
schubev, line 12: I don’t think requiring users to specify `routes.ArgName("arg")` when using the decorator is the way to go : it is very legible and does not provide any new info. Perhaps you should redefine your structures to use a simple `string` and convert them to `ArgName` behind the scene, that way `routes.ParseInt{routes.ArgName("arg")}` becomes `routes.ParseInt{"arg"}`.

routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	ParseEnd struct{}
+)
+
+const (
+	index = parseIndex(iota)
+)
+
+var (
+	badURLTypeMessage  = ErrorBody{"Bad URL type."}
```
schubev, line 37: When you send this, the URL is fine : the type of the argument is unexpected. The name and message should reflect that.

routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
+package routes
+
+import (
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+type (
+	parseIndex uint
```
schubev, line 10: If you want to create an atom/symbol, you should rather use `struct{}`, like so.
```go
type parseIndex struct{}
```
And then use the `parseIndex{}` value.

routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
...
+				a[d.ArgName] = i
+				return h(w, r, a)
+			}
+			return badURLTypeCode, badURLTypeMessage
+		}
+		return notFoundURLCode, notFoundURLMessage
+	}
+}
+
+func (d ParseFloat32) Decorate(h IntermediateHandler) IntermediateHandler {
```
schubev, line 67: I don’t think defining this decorator for floats is necessary, also there is some complexity in handling floats which I think we should avoid : for example your parser allows "NaN", "+Inf", "-Inf", or "-0" which would be passed to a handler which may or may not be able to deal with those.
Unless you have an immediate example of why this is useful, I think you should drop `ParseFloat32` and  `ParseFloat64`.

routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+var (
+	badURLTypeMessage  = ErrorBody{"Bad URL type."}
+	badURLTypeCode     = 400
+	notFoundURLMessage = ErrorBody{"Not Found."}
+	notFoundURLCode    = 404
+)
+
+func incrementIndex(a Arguments) int {
```
schubev, line 43: I was surprised when I realised later in this file that `incrementIndex` returns something : the name doesn’t reflect that at all. You shouldn’t write functions which both _do_ something and _give you_ something unless it is clear what they do and return.

routes/variableurl.go
```diff
@@ -0,0 +1,123 @@
...
+	ParseFloat64 struct {
+		ArgName ArgName
+	}
+	ParseString struct {
+		ArgName ArgName
+	}
+	ParseVoid struct {
+		Required string
+	}
+	ParseEnd struct{}
```
schubev, line 29: Again, you’re not parsing anything here, consider renaming this atom.

server/server.go
```diff
@@ -69,7 +69,8 @@ func initializeHandlers() {
 	}
 }
 
-func applyDecoratorsAndHandle(p string, h routes.IntermediateHandler, ds []routes.Decorator) {
-	h = routes.ApplyDecorators(h, ds...)
+func applyDecoratorsAndHandle(p string, ih routes.IntermediateHandler, ds []routes.Decorator) {
+	ih = routes.ApplyDecorators(ih, ds...)
+	h := http.StripPrefix(p, ih)
```
schubev, line 74: Please do that using a decorator, then use that decorator in `server.go` where there is a list of decorators applied to all routes.

schubev (changes requested):

routes/variableurl.go
schubev: My bad, disregard that one, I thought you were making a new package for a moment.

schubev:
Following our discussion including both the front-end and back-end teams, we decided to use the query string for this kind of arguments. Closing now.

PR #14, opened 2017-11-15
Route to get AWS accounts properties with filters on the IDs.

thibautcornolti:


schubev:
What is the status on this one ? We agreed the “work in progress” tag meant the PR is not ready for review. Is this still the case ?

giubil (dismissed):

giubil (dismissed):

aws/routeAwsGet.go
```diff
@@ -26,15 +26,40 @@ import (
...
+	for _, idd := range slice {
+		if idd == id {
+			return true
+		}
+	}
+	return false
+}
+
+// stripAwsAccounts strips the slice of AwsAccount with filters in the
+// arguments map with QueryArgAwsAccounts as the key.
```
schubev: I can’t understand that sentence.

aws/routes.go
```diff
@@ -27,13 +27,21 @@ import (
 	"github.com/trackit/trackit2/users"
 )
 
+var (
+	// QueryArgAwsAccounts allows to get the AWS Account IDs in the URL Parameters
+	// with routes.RequiredQueryArgs. This AWS Account IDs will be a slice of Uint
+	// stored in the routes.Arguments map with itself for key.
+	QueryArgAwsAccounts = routes.QueryArg{"aa", "AWS Account IDs", routes.QueryArgUintSlice{}, false}
```
schubev: Why don’t you put this definition next to that of `QueryArgAwsAccount` ? Please keep similar definitions in the same place.

routes/decoratorQueryArgs.go
```diff
@@ -61,11 +61,13 @@ type (
 	}
 
 	// QueryArg defines an argument by its name and its type. A description
-	// can be used for documentation purposes.
+	// can be used for documentation purposes. The argument will be mandatory
+	// if you notice it.
```
schubev:
```
The argument will be mandatory if you notice it.
```
This makes no sense.

routes/decoratorQueryArgs.go
```diff
@@ -61,11 +61,13 @@ type (
...
 	QueryArg struct {
 		Name        string
 		Description string
 		Type        QueryParser
+		Mandatory   bool
```
schubev: Previous behavior was : every `QueryArg` represents a mandatory query argument. Now you add a `Mandatory` field whose default value is `false` (the zero bool value), so all previous uses of `QueryArg` now become optional. Finally, you _did not_ go back on all previous definitions of `QueryArg` to make them mandatory again. You should be very careful with this kind of changes and think about how you change the behavior of what’s already there. An easy fix is to replace your `Mandatory` field by an `Optional` field, which defaulting to `false`, _does not_ change the behavior of previously defined `QueryArgs`. Please do that.
The other solution would be to go back on every definition of a `QueryArg` to add `true` to its definition, which is practical in this situation because the codebase is young, but would be harder with a larger codebase.

aws/routeAwsGet.go
```diff
@@ -26,15 +26,40 @@ import (
...
+func stripAwsAccounts(awsAccounts []AwsAccount, a routes.Arguments) []AwsAccount {
```
schubev: With this method, you still get every account from the database, only to drop them. Please implement this feature with database queries instead. In SQL you can select `WHERE value IN (42, 32, 10);` (or some similar syntax).

schubev (dismissed):
You broke `QueryArg`, and you use SQL inefficiently. The logic is sound, however.

aws/routeAwsGet.go
```diff
@@ -18,21 +18,73 @@ import (
...
 
 	"github.com/trackit/trackit2/db"
+	"github.com/trackit/trackit2/models"
 	"github.com/trackit/trackit2/routes"
 	"github.com/trackit/trackit2/users"
 )
 
+// AwsAccountsFromUserIDByAccountID retrieves rows from 'trackit.aws_account' as AwsAccount.
+// The result is filtered by a slice of accountID
+func AwsAccountsFromUserIDByAccountID(db models.XODB, userID int, accountIDs []int) ([]AwsAccount, error) {
```
schubev, line 34: Ideally this should be done with XO templates. Let’s leave this here and I’ll create a task to track this.

aws/routeAwsGet.go
```diff
@@ -18,21 +18,73 @@ import (
...
+		stringAccountIDs = append(stringAccountIDs, strconv.FormatInt(int64(id), 10))
+	}
+	accountID := "(" + strings.Join(stringAccountIDs, ",") + ")"
+
+	// sql query
+	var sqlstr = `SELECT ` +
+		`id, user_id, pretty, role_arn, external ` +
+		`FROM trackit.aws_account ` +
+		`WHERE user_id = ? ` +
+		`AND id IN ` + accountID
```
schubev, line 49: We currently use MySQL and the driver does not support directly using slices directly here so this can’t get much better.

aws/routes.go
```diff
@@ -34,6 +34,7 @@ func init() {
 				Summary:     "get aws accounts' data",
 				Description: "Gets the data for all of the user's AWS accounts.",
 			},
+			routes.RequiredQueryArgs{AwsAccountsOptionalQueryArg},
```
schubev: The decorator was named `RequiredQueryArgs` because it was expected there would be an `OptionalQueryArgs` to go along with it. Since this PR introduces the `Optional` field, this doesn’t make sense anymore. Please rename this decorator to just `QueryArgs`.

routes/decoratorQueryArgs.go
```diff
@@ -61,11 +61,13 @@ type (
...
 
 	// QueryArg defines an argument by its name and its type. A description
-	// can be used for documentation purposes.
+	// can be used for documentation purposes. The argument can be optional
+	// by changing the Optional value to true.
 	QueryArg struct {
 		Name        string
 		Description string
 		Type        QueryParser
+		Optional    bool
```
schubev, line 72: Decorators also generate documentation. You add this parameter yet you don’t change the documentation generation and optional arguments are still marked as required. Please edit the documentation generation logic to show optional arguments as optional.

routes/decoratorQueryArgs_test.go
```diff
@@ -22,11 +22,11 @@ import (
...
-	QueryArgTestInt       = QueryArg{"testInt", "Test signed integer", QueryArgInt{}}
-	QueryArgTestUint      = QueryArg{"testUint", "Test unsigned integer", QueryArgUint{}}
-	QueryArgTestString    = QueryArg{"testString", "Test string", QueryArgString{}}
-	QueryArgTestIntSlice  = QueryArg{"testIntSlice", "Test signed integer slice", QueryArgIntSlice{}}
-	QueryArgTestUintSlice = QueryArg{"testUintSlice", "Test unsigned integer slice", QueryArgUintSlice{}}
+	QueryArgTestInt       = QueryArg{"testInt", "Test signed integer", QueryArgInt{}, false}
+	QueryArgTestUint      = QueryArg{"testUint", "Test unsigned integer", QueryArgUint{}, false}
+	QueryArgTestString    = QueryArg{"testString", "Test string", QueryArgString{}, false}
+	QueryArgTestIntSlice  = QueryArg{"testIntSlice", "Test signed integer slice", QueryArgIntSlice{}, false}
+	QueryArgTestUintSlice = QueryArg{"testUintSlice", "Test unsigned integer slice", QueryArgUintSlice{}, false}
```
schubev: Please test the logic and generated documentation for optional arguments.

schubev (dismissed):

giubil (approved):

schubev (approved):

PR #16, opened 2017-11-21
Store EC2 instance pricing in DB

thibautcornolti:
Fetch EC2 Pricing thanks to the AWS API and store it in Database.


aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/trackit/jsonlog"
+	"github.com/trackit/trackit2/db"
+	"github.com/trackit/trackit2/models"
+)
+
+var (
+	// storeAttribute contains the reference between
+	// a field name and the way to store its value.
+	storeAttribute = map[string]func(string, interface{}, *models.AwsProductEc2, jsonlog.Logger){
```
giubil: `func(string, interface{}, *models.AwsProductEc2, jsonlog.logger)` is not a descriptive type name, why not alias it with something more descriptive, such as `attributeStoring` ?

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+		"operatingSystem":    storeAttributeGenericString,
+		"ecu":                storeAttributeGenericInt,
+		"location":           storeAttributeRegion,
+	}
+)
+
+const (
+	// urlEC2Pricing is the URL used by downloadJSON to
+	// fetch the EC2 pricing.
+	urlEC2Pricing = "https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/index.json"
```
giubil: You should put this url inside of a config file instead of putting it directly into the code

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+// storeAttributeRegion creates the region in the database
+// if it doesn't exist. Then, the function stores the region's
+// ID in the struct which will be imported in the database.
+func storeAttributeRegion(name string, value interface{}, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger) {
+	dbAwsRegion, err := models.AwsRegionByPretty(db.Db, value.(string))
+	if err != nil {
```
giubil: You check if `err != nil` but does it only mean that the region is not created inside the database ?
You do not check if it could be another type of error

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+		var newDbAwsRegion models.AwsRegion
+		logger.Error("Error during dbAwsRegion fetching", err)
+		newDbAwsRegion.Pretty = value.(string)
+		newDbAwsRegion.Region = value.(string)
+		if err := newDbAwsRegion.Insert(db.Db); err != nil {
+			logger.Error("Error during dbAwsRegion inserting", err)
```
giubil: You log an error but do not return it

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+			case "ecu":
+				dbAwsProduct.Ecu = value
+			}
+		}
+	}
+}
+
+// storeAttributeGenericBool stores an bool value in the struct
+// which will be imported in the database.
+func storeAttributeGenericBool(name string, value interface{}, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger) {
```
giubil: Your function does not return anything.
Please try to avoid this model as it makes for unmaintainable code quickly.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+		switch name {
+		case "currentGeneration":
+			dbAwsProduct.CurrentGeneration = val
+		}
+	}
+}
+
+// storeAttributeGenericString stores an string value in the struct
+// which will be imported in the database.
+func storeAttributeGenericString(name string, value interface{}, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger) {
```
giubil: Ditto here for a function that does not return anything

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+}
+
+// importResult parses the body returned by downloadJSON and
+// inserts the pricing in the database.
+func importResult(reader io.ReadCloser, logger jsonlog.Logger) {
+	defer reader.Close()
+	body, err := ioutil.ReadAll(reader)
+	if err != nil {
+		logger.Error("Body reading failed", err)
```
giubil: You log an error but do not return it or even return from the function

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	var bodyMap map[string]interface{}
+	if err = json.Unmarshal(body, &bodyMap); err != nil {
+		logger.Error("JSON Unmarshal failed", err)
+		return
```
giubil: You log an error but do not return it

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+		var dbAwsProduct models.AwsProductEc2
+		product := product.(map[string]interface{})
+		dbAwsProduct.Sku = product["sku"].(string)
+		attributes := product["attributes"].(map[string]interface{})
+		for name, attribute := range attributes {
+			if fct, ok := storeAttribute[name]; ok {
+				fct(name, attribute, &dbAwsProduct, logger)
+			}
+		}
+		dbAwsProduct.Insert(db.Db)
```
giubil: As stated in the function prototype : `func (ape *AwsProductEc2) Insert(db XODB) error {` this method can return an error but you do not check for it
This also means that this functions must return an error

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// purgeCurrent purges the current pricing from the database.
+func purgeCurrent() {
+	// sql query
+	const sqlstr = `DELETE FROM trackit.aws_product_ec2`
+
+	// run query
+	models.XOLog(sqlstr)
+	db.Db.Exec(sqlstr)
```
giubil: I haven't checked the `mysql` go package, but I am pretty sure that a `Exec` statement can generate an error, but you do not check for it

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+	return res.Body, nil
+}
+
+// ImportEC2Pricing downloads the EC2 Pricing from AWS and
+// store it in the database.
+func ImportEC2Pricing() (interface{}, error) {
+	logger := jsonlog.DefaultLogger
+	res, err := downloadJSON(logger)
+	if err != nil {
+		logger.Error("Error during JSON downloading", err)
```
giubil: You log an error but do not return it

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+	} else if res != nil {
+		purgeCurrent()
+		importResult(res, logger)
+	}
+	return nil, nil
```
giubil: Why do you need the `interface{}` return type, if you will always return `nil` ?

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+}
```
giubil: Please write Unit Tests

giubil (dismissed):

aws/product/awsec2product.go
schubev: In that case, why is there even a return value ?

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+
+	"github.com/trackit/jsonlog"
+	"github.com/trackit/trackit2/db"
+	"github.com/trackit/trackit2/models"
```
schubev: Please respect organize your imports in three sections : standard library imports, imports from other projects (not `github.com/trackit/trackit2`) and imports from the current project.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+const (
+	// urlEC2Pricing is the URL used by downloadJSON to
+	// fetch the EC2 pricing.
+	urlEC2Pricing = "https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/index.json"
+)
+
+// storeAttributeRegion creates the region in the database
+// if it doesn't exist. Then, the function stores the region's
+// ID in the struct which will be imported in the database.
+func storeAttributeRegion(name string, value interface{}, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger) {
```
schubev: What does `value` represent ? Neither the name nor the type mean anything.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+	}
+}
+
+// storeAttributeGenericBool stores an bool value in the struct
+// which will be imported in the database.
+func storeAttributeGenericBool(name string, value interface{}, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger) {
+	if value, ok := value.(string); ok {
+		var val int
+		if "Yes" == value {
```
schubev: Yoda conditions not used we have. By the current style abide you shall.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+			val = 1
+		}
+		switch name {
```
schubev: Why do you add a single-cased `switch` when you just used an `if` for the same purpose ?

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+		case "operatingSystem":
+			dbAwsProduct.OperatingSystem = value
+		}
+	}
+}
+
+// importResult parses the body returned by downloadJSON and
+// inserts the pricing in the database.
+func importResult(reader io.ReadCloser, logger jsonlog.Logger) {
+	defer reader.Close()
```
schubev: Good !

aws/product/awsec2product.go
schubev: Please do your SQL queries only in the `models` package, which abstracts all database access. Since we don’t really have an ORM this is how we abstract storage.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,191 @@
...
+// checking if there's already the same version in the database.
+func downloadJSON(logger jsonlog.Logger) (io.ReadCloser, error) {
+	hc := http.Client{}
+	req, err := http.NewRequest("GET", urlEC2Pricing, nil)
+	if err != nil {
+		return nil, err
+	}
+	lastFetch, err := models.AwsFetchPricingByProduct(db.Db, "ec2")
+	if err != nil {
+		req.Header.Add("If-None-Match", lastFetch.Etag)
```
schubev: It’s good that you make sure we don’t process something we already have processed, though perhaps you should extract this in its own function. This one is very long.

schubev (dismissed):
# JSON parsing
Go’s `encoding/json` is very good at parsing data when you know what it looks like, which you do here. By defining some types you can get `encoding/json` to handle the parsing in a typesafe manner. For example:
```go
type PricingSku string

type Ec2Princing struct {
    PublicationDate time.Time
    Products map[PricingSku]PricingProduct
}

type PricingProduct string {
  Sku PricingSku
  ProductFamily string
  Attributes Ec2ProductAttributes
}

type Ec2ProductAttributes struct {
  InstanceType string
  InstanceFamily string
  Memory string
  Storage string
}
```

# Memory usage
As of this review, the document is 172MiB and your current method stores it all in memory. Please try to use streaming when working with such data. https://golang.org/pkg/encoding/json/#example_Decoder_Decode_stream shows how you can decode a JSON document as a stream without storing it all.

# Lack of transactions
You do not use an SQL transaction for this job. This means that in case of error you will leave the database in a dirty state. Please use a transaction to commit your changes to storage only once the whole job is finished.

# Lack of context
Please use `context.Context`, this allows several things : a logger can be stored there which will add some ID to all log events for correlation ; it can store the SQL transaction ; it lets the job be cancelled cleanly.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+		ProductFamily string
+		Attributes    PricingAttribute
+	}
+
+	EC2Pricing struct {
+		Products map[PricingSku]PricingProduct
+	}
+)
+
+func storeRegionID(region string, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger, sqlTx models.XODB) error {
```
schubev: `sqlTx` is just named `tx` everywhere else in the project. Why change the naming convention ?

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	dbAwsRegion, err := models.AwsRegionByPretty(sqlTx, region)
+	if err != nil && err.Error() == "sql: no rows in result set" {
+		var newDbAwsRegion models.AwsRegion
```
schubev: This is clearer as
```go
var newDbAwsRegion = models.AwsRegion{
    Pretty = region,
    Region = region,
}
```

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+		newDbAwsRegion.Pretty = region
+		newDbAwsRegion.Region = region
+		if err := newDbAwsRegion.Insert(sqlTx); err != nil {
+			logger.Error("Error during dbAwsRegion inserting", err.Error())
```
schubev: If I were to see this in the logs, I’d have no idea what it means. When logging, ensure your message is meaningful outside the code.

aws/product/awsec2product.go
schubev: You modify the arguments while neither documenting the function nor reflecting that fact in the function name. Don’t.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+			return err
+		}
+		dbAwsProduct.RegionID = newDbAwsRegion.ID
+	} else if err == nil {
+		dbAwsProduct.RegionID = dbAwsRegion.ID
+	} else {
+		logger.Error("Error during dbAwsRegion fetching", err.Error())
```
schubev: Same here, I’d have no idea what that means if I saw it pop in my logs.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	EC2Pricing struct {
+		Products map[PricingSku]PricingProduct
+	}
+)
+
+func storeRegionID(region string, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger, sqlTx models.XODB) error {
+	dbAwsRegion, err := models.AwsRegionByPretty(sqlTx, region)
+	if err != nil && err.Error() == "sql: no rows in result set" {
```
schubev: Please directly compare to `sql.ErrNoRows`.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+
+	"github.com/trackit/jsonlog"
+
+	"github.com/trackit/trackit2/config"
+	"github.com/trackit/trackit2/db"
+	"github.com/trackit/trackit2/models"
+)
+
+type (
+	PricingSku string
```
schubev: An SKU is an SKU, you don’t need to say it’s a “Pricing SKU”.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+		return err
+	}
+	return nil
+}
+
+// storeAttributes stores all the attributes from PricingAttribute
+// to models.AwsProductEc2
+func storeAttributes(attributes *PricingAttribute, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger, sqlTx models.XODB) (err error) {
+	dbAwsProduct.InstanceType = attributes.InstanceType
+	if attributes.CurrentGeneration == "Yes" {
```
schubev: Do you expect only the two strings “Yes” and “No” ? Then make sure an error is logged if we get an unexpected string.

docker/sql/schema.sql
```diff
@@ -52,3 +52,38 @@ CREATE TABLE aws_bill_repository (
 CREATE VIEW aws_bill_repository_due_update AS
 	SELECT * FROM aws_bill_repository WHERE next_update <= NOW()
 ;
+
+CREATE TABLE aws_fetch_pricing (
```
schubev: What is an AWS fetch pricing ?

docker/sql/schema.sql
```diff
@@ -52,3 +52,38 @@ CREATE TABLE aws_bill_repository (
...
+	CONSTRAINT UNIQUE KEY (region),
+	CONSTRAINT UNIQUE KEY (pretty)
+);
+
+CREATE TABLE aws_product_ec2 (
+	id INTEGER NOT NULL AUTO_INCREMENT,
+	sku VARCHAR(255) NOT NULL,
+	region_id INTEGER NOT NULL,
+	instance_type VARCHAR(255) NOT NULL,
+	current_generation INTEGER(1) NOT NULL,
```
schubev: Use a boolean.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	EC2Pricing struct {
+		Products map[PricingSku]PricingProduct
+	}
+)
+
+func storeRegionID(region string, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger, sqlTx models.XODB) error {
+	dbAwsRegion, err := models.AwsRegionByPretty(sqlTx, region)
+	if err != nil && err.Error() == "sql: no rows in result set" {
+		var newDbAwsRegion models.AwsRegion
+		newDbAwsRegion.Pretty = region
```
schubev: Why the extra “pretty” field if it’s just the same value ? This seems useless and the sole justification for creating a new table for regions. Adding a join is adding complexity, and it seems we do not need it at all.

config/config.go
```diff
@@ -44,18 +44,87 @@ var (
...
-	flag.StringVar(&SqlProtocol, "sql-protocol", "mysql", "The protocol used to communicate with the SQL database.")
-	flag.StringVar(&SqlAddress, "sql-address", "trackit:trackitpassword@tcp(127.0.0.1)/trackit", "The address (username, password, transport, address and database) for the SQL database.")
-	flag.StringVar(&AuthIssuer, "auth-issuer", "trackit", "The 'iss' field for the JWT tokens.")
-	flag.StringVar(&AuthSecret, "auth-secret", "trackitdefaultsecret", "The secret used to sign and verify JWT tokens.")
-	flag.StringVar(&AwsRegion, "aws-region", "us-east-1", "The AWS region the server operates in.")
-	flag.StringVar(&BackendId, "backend-id", "", "The ID to be sent to clients through the 'X-Backend-ID' field. Generated if left empty.")
-	flag.StringVar(&EsAuthentication, "es-auth", "basic:elastic:changeme", "The authentication to use to connect to the ElasticSearch database.")
-	flag.StringVar(&EsAddress, "es-address", "http://127.0.0.1:9200", "The address of the ElasticSearch database.")
-	flag.BoolVar(&PrettyJsonResponses, "pretty-json-responses", false, "JSON HTTP responses should be pretty.")
```
schubev: No.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	dbAwsProduct.Memory = attributes.Memory
+	dbAwsProduct.Storage = attributes.Storage
+	dbAwsProduct.NetworkPerformance = attributes.NetworkPerformance
+	dbAwsProduct.Tenancy = attributes.Tenancy
+	dbAwsProduct.OperatingSystem = attributes.OperatingSystem
+	if val, err := strconv.Atoi(attributes.Ecu); err == nil {
+		dbAwsProduct.Ecu = val
+	}
+	if attributes.LocationType == "AWS Region" {
```
schubev: Again, please ensure unexpected values, for example for new types of locations, throw errors so that we may be notified the code needs updating.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// storeAttributes stores all the attributes from PricingAttribute
+// to models.AwsProductEc2
+func storeAttributes(attributes *PricingAttribute, dbAwsProduct *models.AwsProductEc2, logger jsonlog.Logger, sqlTx models.XODB) (err error) {
+	dbAwsProduct.InstanceType = attributes.InstanceType
+	if attributes.CurrentGeneration == "Yes" {
+		dbAwsProduct.CurrentGeneration = 1
+	}
+	if val, err := strconv.Atoi(attributes.Vcpu); err == nil {
```
schubev: You should at the very least log this kind of error.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	lastFetch, err := models.AwsFetchPricingByProduct(db.Db, "ec2")
+	if err == nil {
+		req.Header.Add("If-None-Match", lastFetch.Etag)
+	}
+	res, err := hc.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode == 304 {
+		logger.Info("JSON Downloading halted: Already exists with ETag.\n", lastFetch.Etag)
```
schubev: `"JSON download halted:…"`

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+// checking if there's already the same version in the database.
+func downloadJSON(logger jsonlog.Logger, sqlTx models.XODB) (io.ReadCloser, error) {
+	hc := http.Client{}
+	req, err := http.NewRequest("GET", config.UrlEC2Pricing, nil)
+	if err != nil {
+		return nil, err
+	}
+	lastFetch, err := models.AwsFetchPricingByProduct(db.Db, "ec2")
+	if err == nil {
+		req.Header.Add("If-None-Match", lastFetch.Etag)
```
schubev: This is very good !

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+
+// downloadJSON requests AWS' API and returns the body after
+// checking if there's already the same version in the database.
+func downloadJSON(logger jsonlog.Logger, sqlTx models.XODB) (io.ReadCloser, error) {
+	hc := http.Client{}
+	req, err := http.NewRequest("GET", config.UrlEC2Pricing, nil)
+	if err != nil {
+		return nil, err
+	}
+	lastFetch, err := models.AwsFetchPricingByProduct(db.Db, "ec2")
```
schubev: Please extract `"ec2"` into a constant. This is generally good practice, because a typo in a string can be hard to debug, while a typo in an identifier usually immediately causes a compile error.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+		Etag:    newEtag,
+	}
+	return dbAfp.Save(sqlTx)
+}
+
+// downloadJSON requests AWS' API and returns the body after
+// checking if there's already the same version in the database.
+func downloadJSON(logger jsonlog.Logger, sqlTx models.XODB) (io.ReadCloser, error) {
+	hc := http.Client{}
+	req, err := http.NewRequest("GET", config.UrlEC2Pricing, nil)
```
schubev: Please use `http.MethodGet` instead of a literal string.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	lastFetch, err := models.AwsFetchPricingByProduct(db.Db, "ec2")
+	if err == nil {
+		req.Header.Add("If-None-Match", lastFetch.Etag)
+	}
+	res, err := hc.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode == 304 {
```
schubev: Please use `http.StatusNotModified`, not everyone know HTTP codes at the top of their head.

schubev: Also, see my comment about the `"ec2"` string, it applies here as well.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+		logger.Info("JSON Downloading halted: Already exists with ETag.\n", lastFetch.Etag)
+		return nil, nil
+	}
+	if err = saveLastFetch(lastFetch, res.Header["Etag"][0], sqlTx); err != nil {
+		logger.Error("Error when saving the dbAfp", err.Error())
```
schubev: Again, ensure your logs are meaningful outside of the code that produce them.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, err
+	}
+	return res.Body, nil
+}
+
+// ImportEC2Pricing downloads the EC2 Pricing from AWS and
+// store it in the database.
+func ImportEC2Pricing() error {
+	logger := jsonlog.DefaultLogger
```
schubev: Please use a context and get your logger from there.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	sqlTx, err := db.Db.BeginTx(context.Background(), nil)
```
schubev: The context given to this function should also be used here instead of the background context.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		logger.Error("Error when beginning sqlTx", err.Error())
```
schubev: Ensure your logs are meaningful without the code that produced them. I won’t make this comment again, check all your logs.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+	hc := http.Client{}
+	req, err := http.NewRequest("GET", config.UrlEC2Pricing, nil)
+	if err != nil {
+		return nil, err
+	}
+	lastFetch, err := models.AwsFetchPricingByProduct(db.Db, "ec2")
+	if err == nil {
+		req.Header.Add("If-None-Match", lastFetch.Etag)
+	}
+	res, err := hc.Do(req)
```
schubev: Use a context.

config/config.go
```diff
@@ -44,18 +44,87 @@ var (
...
+		"pretty-json-responses",
+		false,
+		"JSON HTTP responses should be pretty.",
+	)
+
+	flag.StringVar(
+		&UrlEC2Pricing,
+		"url-ec2-pricing",
+		"https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/index.json",
+		"The URL used to download the EC2 pricing",
```
schubev: You forgot the dot.

docker/sql/schema.sql
```diff
@@ -52,3 +52,38 @@ CREATE TABLE aws_bill_repository (
...
+
+CREATE TABLE aws_fetch_pricing (
+	id INTEGER NOT NULL AUTO_INCREMENT,
+	product VARCHAR(255) NOT NULL,
+	etag VARCHAR(255) NOT NULL,
+	CONSTRAINT PRIMARY KEY (id),
+	CONSTRAINT UNIQUE KEY (product)
+);
+
+CREATE TABLE aws_region (
```
schubev: Because you just store the same data in `region` and `pretty`, this table adds nothing and increases complexity. I think you should have a look at “Codd’s normal forms”, BTW.

docker/sql/schema.sql
```diff
@@ -52,3 +52,38 @@ CREATE TABLE aws_bill_repository (
...
+	current_generation INTEGER(1) NOT NULL,
+	vcpu INTEGER NOT NULL,
+	memory VARCHAR(255) NOT NULL,
+	storage VARCHAR(255) NOT NULL,
+	network_performance VARCHAR(255) NOT NULL,
+	tenancy VARCHAR(255) NOT NULL,
+	operating_system VARCHAR(255) NOT NULL,
+	ecu INTEGER NOT NULL,
+	CONSTRAINT PRIMARY KEY (id),
+	CONSTRAINT UNIQUE KEY (sku),
```
schubev: Why not use the SKU as a primary key ? You can use strings as primary keys. Also, if you can find documentation that SKU’s have a fixed, number of characters, you can even use `CHAR` instead of `VARCHAR`.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// importResult parses the body returned by downloadJSON and
+// inserts the pricing in the database.
+func importResult(reader io.ReadCloser, logger jsonlog.Logger, sqlTx models.XODB) error {
+	defer reader.Close()
+
+	decoder := json.NewDecoder(reader)
+	var bodyMap EC2Pricing
+	if err := decoder.Decode(&bodyMap); err != nil {
```
schubev: Please stream the JSON document as per my previous review. This document is known to be quite big and reasonable memory footprint is a requirement for this project.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,179 @@
...
+// ImportEC2Pricing downloads the EC2 Pricing from AWS and
+// store it in the database.
+func ImportEC2Pricing() error {
+	logger := jsonlog.DefaultLogger
+	sqlTx, err := db.Db.BeginTx(context.Background(), nil)
+	if err != nil {
+		logger.Error("Error when beginning sqlTx", err.Error())
+		return err
+	}
+	res, err := downloadJSON(logger, sqlTx)
```
schubev: Here, if there are no errors, you
1. Open a connection to download a JSON document
2. Purge a table in the database
3. Consume the JSON document

You can’t know for sure that the database purge will end before the connection times out, so you have two possible solutions :
1. Insert products along with the etag (or some other id) for the document they came from, with an `INSERT … ON DUPLICATE KEY UPDATE`, then remove from the database all records with another etag (this is my preferred option).
2. Immediately consume the JSON document by writing it to a temporary file, purge the database, then consume the temporary file (don’t forget to remove it afterwards).

schubev (dismissed):
You handled downloading the JSON document well by checking the ETag and errors.

However, in my previous I asked you to consume the JSON document as a stream, which you didn’t.

Also, please use contexts wherever you perform long-running, blocking actions such as database requests (you can attach the context to the transaction) or HTTP requests. Contexts allow jobs to be cancelled, and carry values useful for logging.

Finally, with SQL you can insert more than one record at a time. You perform an SQL request for each and every SKU, which means an additional network roundtrip everytime. This can lead to poor performances. Please batch your inserts when they are numerous.

aws/product/awsec2product.go
```diff
@@ -0,0 +1,226 @@
...
+		ProductFamily string
+		Attributes    Attribute
+	}
+)
+
+// storeAttributes stores all the attributes from Attribute
+// to models.AwsProductEc2
+func storeAttributes(ctx context.Context, attributes *Attribute,
+	dbAwsProductPricing *models.AwsProductPricingEc2, tx models.XODB) {
+	logger := jsonlog.LoggerFromContextOrDefault(ctx)
```
giubil: You take `tx models.XODB` as a parameter, but you do not use it.

giubil (dismissed):

schubev (approved):
The SQL insert part is clunky, but with the tools we currently use I don’t really know how we can do better.

PR #18, opened 2017-11-27
Monitor EC2 instance stats

thibautcornolti:
Add a way to monitor the EC2 instances and store the informations in ElasticSearch.


aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+
+	"github.com/trackit/trackit2/config"
+	"github.com/trackit/trackit2/es"
+)
+
+const (
+	defaultRegion      = "us-west-2"
+	elasticsearchIndex = "awshourlyinstanceusage"
+	elasticsearchType  = "ec2instance"
+)
```
giubil: Should this 3 const be here, or should they be in a conf file ?
Since this project is open source, people might need to change those, it would be easier to have them in a config file.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	sess := session.Must(session.NewSession(&aws.Config{
+		Credentials: creds,
+		Region:      aws.String(defaultRegion),
+	}))
+	svc := sts.New(sess)
+	res, err := svc.GetCallerIdentity(nil)
+	if err != nil {
+		logger.Error("Get caller identity failed", err)
+		return ""
+	}
```
giubil: This logs an error but doesn't return one, which doesn't really make sense

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	return aws.StringValue(res.Account)
+}
+
+// getInstanceTag formats an array of *ec2.Tag to an array
+// of InstanceTag.
+func getInstanceTag(tags []*ec2.Tag) []InstanceTag {
+	res := make([]InstanceTag, 0, 5)
```
giubil: Why use `5` in the make if you append to it ?

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	defer close(instanceInfoChan)
+	sess := session.Must(session.NewSession(&aws.Config{
+		Credentials: creds,
+		Region:      aws.String(region),
+	}))
+	svc := ec2.New(sess)
+	instances, err := svc.DescribeInstances(nil)
+	if err != nil {
+		logger.Error("Describe instances failed", err)
+		return
```
giubil: Again you log an error but you don't return one

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+func fetchRegionsList(creds *credentials.Credentials, logger jsonlog.Logger) []string {
+	sess := session.Must(session.NewSession(&aws.Config{
+		Credentials: creds,
+		Region:      aws.String(defaultRegion),
+	}))
+	svc := ec2.New(sess)
+	regions, err := svc.DescribeRegions(nil)
+	if err != nil {
+		logger.Error("Describe regions failed", err)
+		return []string{}
```
giubil: Again you log an error but don't return one

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	res := make([]string, 0, 10)
```
giubil: Why use `10` in the make if you append to it ?

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	for _, region := range regions.Regions {
+		res = append(res, aws.StringValue(region.RegionName))
+	}
+	return res
+}
+
+// getPropertiesMapping uses the reflection of the struct InstanceInfo and
+// InstanceTag to get the mapping of the ElasticSearch documents.
+func getPropertiesMapping(field interface{}) map[string]interface{} {
```
giubil: A function that takes `field interface{}` and returns `map[string]interface{}` is really not explicit.
Either stop using `interface{}` or explain them in the documentation

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	properties := getPropertiesMapping(InstanceInfo{})
+	mapping := map[string]interface{}{
+		"mappings": map[string]interface{}{
+			elasticsearchType: map[string]interface{}{
+				"properties": properties,
+			},
+		},
+	}
+	if _, err := client.CreateIndex(elasticsearchIndex).BodyJson(mapping).Do(context.Background()); err != nil {
+		logger.Error("ElasticSearch create index failed", err)
```
giubil: Again you log an error but don't return one

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+		Namespace:  aws.String("AWS/EC2"),
+		MetricName: aws.String("CPUUtilization"),
+		StartTime:  aws.Time(start),
+		EndTime:    aws.Time(end),
+		Period:     aws.Int64(int64(86400 * 30)),
+		Statistics: []*string{aws.String("Average"), aws.String("Maximum")},
+		Dimensions: dimensions,
+	})
+	if err != nil {
+		logger.Error("Fetch CPU stats from CloudWatch failed", err)
```
giubil: Again you log an error but don't return one

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	if _, err := client.CreateIndex(elasticsearchIndex).BodyJson(mapping).Do(context.Background()); err != nil {
+		logger.Error("ElasticSearch create index failed", err)
+		return
+	}
+}
+
+// importInstancesToES imports an array of InstanceInfo in ElasticSearch.
+// It calls createIndexES if the index doesn't exist.
+func importInstancesToES(instances []InstanceInfo, logger jsonlog.Logger) {
+	// TODO: Get credentials
```
giubil: Why create a PR if you still have `TODO` comments ?

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	client, err := es.NewSignedElasticClient(config.EsUrl, credentials.NewSharedCredentials("", "default"))
+	if err != nil {
+		logger.Error("Get ElasticSearch client failed", err)
```
giubil: Again you log an error but don't return one

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+		return
+	}
+	if exists, err := client.IndexExists(elasticsearchIndex).Do(context.Background()); err != nil {
+		logger.Error("ElasticSearch check if index exists failed", err)
```
giubil: Again you log an error but don't return one

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+		return
+	} else if !exists {
+		createIndexES(client, logger)
+	}
+	for _, instance := range instances {
+		if res, err := client.Index().Index(elasticsearchIndex).Type(elasticsearchType).BodyJson(instance).Do(context.Background()); err != nil {
+			logger.Error("Instance put in ES failed", err)
```
giubil: Again you log an error but don't return one

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+		return
+	}
+	regions := fetchRegionsList(creds, logger)
+	instanceInfoChans := make([]<-chan InstanceInfo, 0, 15)
+	for _, region := range regions {
+		instanceInfoChan := make(chan InstanceInfo)
+		go fetchInstancesList(creds, region, account, logger, instanceInfoChan)
+		instanceInfoChans = append(instanceInfoChans, instanceInfoChan)
+	}
+	instances := make([]InstanceInfo, 0, 20)
```
giubil: Why use `20` in the make if you append to it ?

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	logger := jsonlog.DefaultLogger
+	tcreds, err := GetTemporaryCredentials(context.Background(), awsAccount, "fetchInstanceStats")
+	creds := credentials.NewStaticCredentials(
+		aws.StringValue(tcreds.Credentials.AccessKeyId),
+		aws.StringValue(tcreds.Credentials.SecretAccessKey),
+		aws.StringValue(tcreds.Credentials.SessionToken),
+	)
+	account := getAccountID(creds, logger)
+	if err != nil {
+		logger.Error("Get temporary credentials failed", err)
```
giubil: Again you log an error but don't return one

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+		instanceInfoChan := make(chan InstanceInfo)
+		go fetchInstancesList(creds, region, account, logger, instanceInfoChan)
+		instanceInfoChans = append(instanceInfoChans, instanceInfoChan)
+	}
+	instances := make([]InstanceInfo, 0, 20)
+	for instance := range merge(instanceInfoChans...) {
+		instances = append(instances, instance)
+	}
+	importInstancesToES(instances, logger)
+}
```
giubil: Please also write unit tests for your code

giubil (dismissed):

aws/monitorinstances.go
schubev: `defaultRegion` definitely shouldn’t be a `const`. Also it already exists in the `config` module I believe. The other two are fine by me.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+const (
+	defaultRegion      = "us-west-2"
+	elasticsearchIndex = "awshourlyinstanceusage"
+	elasticsearchType  = "ec2instance"
+)
+
+var (
+	// tagsAvailable is used to reflect the structs
+	// InstanceTag and InstanceInfo.
+	tagsAvailable = []string{"type", "index"}
```
schubev: Why use a slice instead of two constants ?

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	// be imported in ElasticSearch.
+	InstanceTag struct {
+		Key   string `json:"key" type:"string" index:"not_analyzed"`
+		Value string `json:"value" type:"string" index:"not_analyzed"`
+	}
+
+	// InstanceInfo represents all the informations of an EC2 instance.
+	// It will be imported in ElasticSearch thanks to the struct tags.
+	InstanceInfo struct {
+		Account    string        `json:"account" type:"string" index:"not_analyzed"`
```
schubev: AFAIK type `string` is deprecated in ES 5.5, instead use type `keyword`. Please check about `not_analyzed`, too.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	// done. This must start after the wg.Add call.
+	go func() {
+		wg.Wait()
+		close(out)
+	}()
+	return out
+}
+
+// getAccountID allows to get the AWS Account ID (12 digits) of the
+// credentials given.
```
schubev: That’s french with english words. How about this ?
```go
// getAccountId gets the AWS Account ID for the given credentials
```

aws/monitorinstances.go
schubev: Indeed, the idiomatic way to do this in Go is to return `(string, error)` where the `string` should be disregarded if the `error` is not nil.

aws/monitorinstances.go
schubev: Also, all other names we’ve written which contain acronyms treat them like normal words, so please name this `getAccountId`.

aws/monitorinstances.go
schubev: It’s a micro-optimization to prevent some allocations, but it’s very minimal and makes the code more complex. I don’t think it should be kept in that case.

aws/monitorinstances.go
schubev: Preferably : don’t use them.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+		properties["tags"] = map[string]interface{}{
+			"properties": getPropertiesMapping(InstanceTag{}),
+		}
+	}
+	return properties
+}
+
+// createIndexES creates the ElasticSearch index with the properties
+// from getPropertiesMapping.
+func createIndexES(client *elastic.Client, logger jsonlog.Logger) {
```
schubev: The code shouldn’t create the mappings. Instead write a template to be provided to the ElasticSearch before Trackit is started.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	if _, err := client.CreateIndex(elasticsearchIndex).BodyJson(mapping).Do(context.Background()); err != nil {
+		logger.Error("ElasticSearch create index failed", err)
+		return
+	}
+}
+
+// importInstancesToES imports an array of InstanceInfo in ElasticSearch.
+// It calls createIndexES if the index doesn't exist.
+func importInstancesToES(instances []InstanceInfo, logger jsonlog.Logger) {
```
schubev: Same thing as earlier about the acronyms in identifiers.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	// TODO: Get credentials
+	client, err := es.NewSignedElasticClient(config.EsUrl, credentials.NewSharedCredentials("", "default"))
```
schubev: You shouldn’t create your own `elastic.Client` when module `es` provides one already.

schubev: Especially since this one is not properly configured : it will work neither in a development environment on a non-AWS elasticsearch, nor on AWS EC2 using EC2 roles.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+		Account    string        `json:"account" type:"string" index:"not_analyzed"`
+		Service    string        `json:"service" type:"string" index:"not_analyzed"`
+		ID         string        `json:"id" type:"string" index:"not_analyzed"`
+		Region     string        `json:"region" type:"string" index:"not_analyzed"`
+		Hour       time.Time     `json:"hour" type:"date" format:"strict_date_optional_time||epoch_millis"`
+		CPUAverage float64       `json:"cpuaverage" type:"double"`
+		CPUPeak    float64       `json:"cpupeak" type:"double"`
+		KeyPair    string        `json:"keypair" type:"string" index:"not_analyzed"`
+		Type       string        `json:"type" type:"string" index:"not_analyzed"`
+		Tags       []InstanceTag `json:"tags" type:"nested"`
```
schubev: ElasticSearch documents representing AWS objects with tags should use the following scheme :
```json
{
    "tags": {
        "foo": "foovalue",
        "bar": "barvalue",
    }
}
```
What you are doing is this.
```json
{
    "tags": [
        {
            "key": "foo",
            "value": "foovalue"
        },
        {
            "key": "bar",
            "value": "barvalue"
        }
    ]
}
```

Use a `map[string]string` instead and ensure your mapping templates includes a dynamic mapping so these are of `keyword` type in ElasticSearch.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	end := time.Date(now.Year(), now.Month(), now.Day(), now.Hour()-1, 0, 0, 0, now.Location())
+	start := time.Date(now.Year(), now.Month(), now.Day(), now.Hour()-2, 0, 0, 0, now.Location())
+	svc := cloudwatch.New(sess)
+	dimensions := []*cloudwatch.Dimension{
+		&cloudwatch.Dimension{
+			Name:  aws.String("InstanceId"),
+			Value: aws.String(instanceId),
+		},
+	}
+	stats, err := svc.GetMetricStatistics(&cloudwatch.GetMetricStatisticsInput{
```
schubev: Please use `context.Context` to
— provide cancelation of long, blocking operations like network communications
— log stuff (please log more)

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+			logger.Info("Instance put in ES", *res)
+		}
+	}
+}
+
+// FetchInstancesStats fetchs the stats of the EC2 instances of an AwsAccount
+// to import them in ElasticSearch. The stats are fetched from the last hour.
+// In this way, FetchInstancesStats should be called every hour.
+func FetchInstancesStats(awsAccount AwsAccount) {
+	logger := jsonlog.DefaultLogger
```
schubev: Please use a `context.Context` and get your logger from there.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	tcreds, err := GetTemporaryCredentials(context.Background(), awsAccount, "fetchInstanceStats")
+	creds := credentials.NewStaticCredentials(
```
schubev: Using STS credentials this way means your function will start failing after one hour. Please look at function `getServiceForRepository` in file _/aws/s3/readBills.go_ for a correct example.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+				CPUAverage: CPUAverage,
+				CPUPeak:    CPUPeak,
+			}
+		}
+	}
+}
+
+// fetchRegionsList fetchs the regions list from AWS and returns an array of their name.
+func fetchRegionsList(creds *credentials.Credentials, logger jsonlog.Logger) []string {
+	sess := session.Must(session.NewSession(&aws.Config{
```
schubev: `session.Session` is safe to use concurrently and it is better not to create one each time you need to use AWS. Please create your session and EC2 service in `FetchInstanceStats` and pass the service around instead of recreating it everywhere.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+		Region:      aws.String(region),
+	}))
+	svc := ec2.New(sess)
+	instances, err := svc.DescribeInstances(nil)
+	if err != nil {
+		logger.Error("Describe instances failed", err)
+		return
+	}
+	now := time.Now()
+	hour := time.Date(now.Year(), now.Month(), now.Day(), now.Hour()-1, 0, 0, 0, now.Location())
```
schubev: Please extract that in a documented function.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+	for _, tag := range tags {
+		res = append(res, InstanceTag{aws.StringValue(tag.Key), aws.StringValue(tag.Value)})
+	}
+	return res
+}
+
+// getInstanceStats gets the CPU average and the CPU peak from CloudWatch
+func getInstanceStats(instanceId string, sess *session.Session, region string, logger jsonlog.Logger) (float64, float64) {
+	now := time.Now()
+	end := time.Date(now.Year(), now.Month(), now.Day(), now.Hour()-1, 0, 0, 0, now.Location())
```
schubev: Please try and find a better way to do that, and otherwise extract it in its own documented function.

aws/monitorinstances.go
```diff
@@ -0,0 +1,308 @@
...
+func getInstanceTag(tags []*ec2.Tag) []InstanceTag {
+	res := make([]InstanceTag, 0, 5)
+	for _, tag := range tags {
+		res = append(res, InstanceTag{aws.StringValue(tag.Key), aws.StringValue(tag.Value)})
+	}
+	return res
+}
+
+// getInstanceStats gets the CPU average and the CPU peak from CloudWatch
+func getInstanceStats(instanceId string, sess *session.Session, region string, logger jsonlog.Logger) (float64, float64) {
```
schubev: Don’t pass a logger, pass a context.

schubev (dismissed):

aws/monitorinstances.go
thibautcornolti: Replaced by `keyword` and `text`. `keyword` is equal to `string not_analyzed` and `text` tokenize the field.

aws/monitorinstances.go
thibautcornolti: It was to loop on it. In that way you could add tags without change the code (except this and the struct).
But it's removed, I changed the way.

aws/ec2/monitorInstances.go
```diff
@@ -0,0 +1,284 @@
...
+			Name:  aws.String("InstanceId"),
+			Value: aws.String(instanceId),
+		},
+	}
+	stats, err := svc.GetMetricStatistics(&cloudwatch.GetMetricStatisticsInput{
+		Namespace:  aws.String("AWS/EC2"),
+		MetricName: aws.String("CPUUtilization"),
+		StartTime:  aws.Time(start),
+		EndTime:    aws.Time(end),
+		Period:     aws.Int64(int64(86400 * 30)),
```
giubil: `86400 * 30` seems like a magic number. You should put it as a constant

aws/ec2/monitorInstances.go
```diff
@@ -0,0 +1,284 @@
...
+		logger.Error("Error when fetching CPU stats from CloudWatch", err.Error())
+		return 0, 0
+	} else if len(stats.Datapoints) > 0 {
+		return aws.Float64Value(stats.Datapoints[0].Average), aws.Float64Value(stats.Datapoints[0].Maximum)
+	} else {
+		return 0, 0
+	}
+}
+
+// fetchInstancesList, firstly, calls DescribeInstances from AWS to list all the instances in a region.
```
giubil: This type of comment does not really follow the coding style for Go comments. The first line should be a description of the function. Here the first line is not a complete description of what the function does

aws/ec2/monitorInstances.go
```diff
@@ -0,0 +1,284 @@
...
+	logger := jsonlog.LoggerFromContextOrDefault(ctx)
+	sess := session.Must(session.NewSession(&aws.Config{
+		Credentials: creds,
+		Region:      aws.String(region),
+	}))
+	svc := ec2.New(sess)
+	instances, err := svc.DescribeInstances(nil)
+	if err != nil {
+		logger.Error("Error when describing instances", err.Error())
+		return
```
giubil: You log an error but don't return one.

aws/ec2/monitorInstances.go
```diff
@@ -0,0 +1,284 @@
...
+	return res, nil
+}
+
+// createIndexEs creates the ElasticSearch index with the properties
+// from getPropertiesMapping.
+func createIndexEs(ctx context.Context, client *elastic.Client) {
+	logger := jsonlog.LoggerFromContextOrDefault(ctx)
+	mapping := "{\"mappings\":{\"" + DocumentEc2 + "\":" + mappingEc2Instance + "}}"
+	if _, err := client.CreateIndex(IndexHourlyInstanceUsage).BodyString(mapping).Do(context.Background()); err != nil {
+		logger.Error("Error when creating ElasticSearch index", err.Error())
```
giubil: If your function can have an error, it should return one. How will the calling function know if this index creation has succeeded ? Should you continue the execution of your calling function if this function fails ?

aws/ec2/monitorInstances.go
```diff
@@ -0,0 +1,284 @@
...
+			Index(IndexHourlyInstanceUsage).
+			Type(DocumentEc2).
+			BodyJson(instance).
+			Id(hash64).
+			Do(context.Background()); err != nil {
+			logger.Error("Error when putting InstanceInfo in ES", err.Error())
+		} else {
+			logger.Info("Instance put in ES", *res)
+		}
+	}
```
giubil, line 247: This function does again not return anything while function called inside of it can return errors.
Since you assume that this function will always succeed, what happens and it fails silently ? Will this be a problem for other functions depending on it ?

aws/ec2/mappings.go
```diff
@@ -0,0 +1,61 @@
...
+		"tags": {
+			"type": "nested"
+		}
+	},
+	"_all": {
+		"enabled": false
+	},
+	"numeric_detection": false,
+	"date_detection": false
+}`
```
giubil, line 61: The mapping is already being inserted inside of the ES via the script `./scripts/add_es_mapping.sh`. What is the purpose of this mapping ?

giubil (dismissed):

aws/ec2/mappings.go
thibautcornolti: It's used to automatically create the index if it doesn't exist. I was inspired by the work of @schubev on the PR#19

PR #23, opened 2017-11-29
Updated golang version in Dockerfile

giubil:
Updated golang version to accommodate `>=1.9` only syntax in the code


schubev (approved):
Wow!

PR #26, opened 2017-12-04
Generate documentation for request bodies

schubev:
The `RequestBody` decorator parses a request body and performs basic sanity checks based on the struct tags. It also generates documentation for the format of those input documents.

### DIFF
--- a/aws/routes.go
+++ b/aws/routes.go
@@ -37,6 +37,11 @@ func init() {
 		),
 		http.MethodPost: routes.H(postAwsAccount).With(
 			routes.RequestContentType{"application/json"},
+			routes.RequestBody{postAwsAccountRequestBody{
+				RoleArn:  "arn:aws:iam::123456789012:role/example",
+				External: "LlzrwHeiM-SGKRLPgaGbeucx_CJC@QBl,_vOEF@o",
+				Pretty:   "My AWS account",
+			}},
 			routes.Documentation{
 				Summary:     "add an aws account",
 				Description: "Adds an AWS account to the user's list of accounts, validating it before succeeding.",
--- a/routes/decoratorRequestBody.go
+++ b/routes/decoratorRequestBody.go
@@ -12,27 +12,27 @@ import (
 	"github.com/trackit/trackit2/util/req"
 )
 
-type JsonRequestBody struct {
+type RequestBody struct {
 	Example interface{}
 }
 
-func (jrb JsonRequestBody) Decorate(h Handler) Handler {
-	h.Func = jrb.getFunc(h.Func)
-	h.Documentation = jrb.getDocumentation(h.Documentation)
+func (rb RequestBody) Decorate(h Handler) Handler {
+	h.Func = rb.getFunc(h.Func)
+	h.Documentation = rb.getDocumentation(h.Documentation)
 	return h
 }
 
-func (jrb JsonRequestBody) getFunc(hf HandlerFunc) HandlerFunc {
-	validate, err := req.CreateValidator(jrb.Example)
+func (rb RequestBody) getFunc(hf HandlerFunc) HandlerFunc {
+	validate, err := req.CreateValidator(rb.Example)
 	var handleWithValidation func(http.ResponseWriter, *http.Request, Arguments, reflect.Value) (int, interface{})
 	if err != nil {
 		logger := jsonlog.DefaultLogger
-		logger.Error("Failed to build validator for type %T.", jrb.Example)
+		logger.Error("Failed to build validator for type %T.", rb.Example)
 		os.Exit(1)
 	}
 	if validate == nil {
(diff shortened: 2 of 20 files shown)


aws/routes.go
```diff
@@ -37,6 +37,11 @@ func init() {
 		),
 		http.MethodPost: routes.H(postAwsAccount).With(
 			routes.RequestContentType{"application/json"},
+			routes.RequestBody{postAwsAccountRequestBody{
+				RoleArn:  "arn:aws:iam::123456789012:role/example",
+				External: "LlzrwHeiM-SGKRLPgaGbeucx_CJC@QBl,_vOEF@o",
```
giubil, line 42: Is this external random ?

routes/decoratorRequestBody.go
```diff
@@ -0,0 +1,133 @@
...
+	return h
+}
+
+func (rb RequestBody) getFunc(hf HandlerFunc) HandlerFunc {
+	validate, err := req.CreateValidator(rb.Example)
+	var handleWithValidation func(http.ResponseWriter, *http.Request, Arguments, reflect.Value) (int, interface{})
+	if err != nil {
+		logger := jsonlog.DefaultLogger
+		logger.Error("Failed to build validator for type %T.", rb.Example)
+		os.Exit(1)
```
giubil, line 32: If there is a `os.Exit(1)` shouldn't the logging be `Fatal` instead of just an `Error` ?

routes/decoratorRequestBody_test.go
```diff
@@ -0,0 +1,97 @@
...
+		t.Errorf("Response should be %v, is %v.", example, response)
+	}
+}
+
+func TestSuccessfulValidation(t *testing.T) {
+	example := bodyEchoTest{"test", 42}
+	handler := H(bodyEchoHandler).With(
+		RequestBody{example},
+	)
+	bs, _ := json.Marshal(example)
```
giubil, line 41: Why don't you check for the `error` return of `json.Marshal` ?

routes/decoratorRequestBody_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+}
+
+func TestFailedValidation(t *testing.T) {
+	const expectedError = "foo: value is zero"
+	example := bodyEchoTest{"", 42}
+	handler := H(bodyEchoHandler).With(
+		RequestBody{example},
+	)
+	bs, _ := json.Marshal(example)
```
giubil, line 59: Ditto here for the `error` value

util/req/req.go
```diff
@@ -0,0 +1,3 @@
+// package req provides request body validation facilities. Using struct tags
+// you may define rules in order to validate structured user input.
+package req
```
giubil, line 3: Why is this file empty ?

util/req/reqSchema.go
```diff
@@ -0,0 +1,82 @@
...
+	"io"
+	"reflect"
+	"strings"
+)
+
+const (
+	schemaIndent = "\t"
+)
+
+func GetSchema(dst io.Writer, typ reflect.Type) error {
```
giubil: The variable name `dst` does not make sense just like this. You need to add a comment to make it more explicit

giubil (changes requested):

aws/routes.go
schubev: It is an example.

routes/decoratorRequestBody.go
schubev: `jsonlog` doesn’t have `Fatal` yet.

util/req/req.go
schubev: I didn’t know in which file to put the package documentation, so I put it here, in a file eponymous to the package, which will not contain code.

util/req/reqSchema.go
schubev: `dst` is quite standard an abbreviation, and with an `io.Writer` seems fairly obvious, but I did forget to document the function.

routes/decoratorRequestBody_test.go
schubev: Because with `example` being a very well-defined structure which `encoding/json` will not fail with, it would only make the _test_ harder to read, with no gain.

routes/decoratorRequestBody_test.go
schubev: Ditto.

routes/decoratorRequestBody.go
giubil: Okay

aws/routes.go
giubil: Okay, that was just to make sure (the `roleArn` looked like an example, but I wasn't sure about the external)

giubil (approved):

PR #27, opened 2017-12-11
Edit an AWS Account

thibautcornolti:
Added PATCH method to the /aws route to edit an AWS Account. Only the pretty field can be edited.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -137,6 +137,20 @@ func (a *AwsAccount) CreateAwsAccount(ctx context.Context, db models.XODB) error
 	return err
 }
 
+// UpdateAWSAccount updates an AWS account for a user. It does no error
+// checking: the caller should check themselves that the AWS account exists.
+// Only the Pretty will be updated.
+func (a *AwsAccount) UpdateAWSAccount(ctx context.Context, tx *sql.Tx) error {
+	logger := jsonlog.LoggerFromContextOrDefault(ctx)
+	dbAwsAccount, _ := models.AwsAccountByID(tx, a.Id)
+	dbAwsAccount.Pretty = a.Pretty
+	err := dbAwsAccount.Update(tx)
+	if err != nil {
+		logger.Error("Failed to update AWS account in database.", nil)
+	}
+	return err
+}
+
 // awsAccountFromDbAwsAccount constructs an aws.AwsAccount from a
 // models.AwsAccount. The distinction exists to decouple database access from
 // the logic of the server.
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -137,19 +137,19 @@ func (a *AwsAccount) CreateAwsAccount(ctx context.Context, db models.XODB) error
 	return err
 }
 
-// UpdateAWSAccount updates an AWS account for a user. It does no error
+// UpdatePrettyAWSAccount updates an AWS account for a user. It does no error
 // checking: the caller should check themselves that the AWS account exists.
 // Only the Pretty will be updated.
-func (a *AwsAccount) UpdateAWSAccount(ctx context.Context, tx *sql.Tx) error {
+func (a *AwsAccount) UpdatePrettyAWSAccount(ctx context.Context, tx *sql.Tx) error {
 	logger := jsonlog.LoggerFromContextOrDefault(ctx)
 	dbAwsAccount, err := models.AwsAccountByID(tx, a.Id)
 	if err != nil {
-		logger.Error("Failed to get AWS account in database.", nil)
+		logger.Error("Failed to get AWS account in database.", err.Error())
 	} else {
 		dbAwsAccount.Pretty = a.Pretty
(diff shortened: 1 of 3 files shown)


aws/aws.go
```diff
@@ -137,6 +137,20 @@ func (a *AwsAccount) CreateAwsAccount(ctx context.Context, db models.XODB) error
 	return err
 }
 
+// UpdateAWSAccount updates an AWS account for a user. It does no error
+// checking: the caller should check themselves that the AWS account exists.
+// Only the Pretty will be updated.
+func (a *AwsAccount) UpdateAWSAccount(ctx context.Context, tx *sql.Tx) error {
+	logger := jsonlog.LoggerFromContextOrDefault(ctx)
+	dbAwsAccount, _ := models.AwsAccountByID(tx, a.Id)
```
giubil: Why don't you check the error return value of `models.AwsAccountByID()` ?

aws/routeAwsPatch.go
```diff
@@ -0,0 +1,70 @@
...
+// patchAwsAccountWithValidBody handles the logic of patchAwsAccount assuming the
+// request body is valid.
+func patchAwsAccountWithValidBody(r *http.Request, tx *sql.Tx, user users.User, body patchAwsAccountRequestBody, id int) (int, interface{}) {
+	ctx := r.Context()
+	logger := jsonlog.LoggerFromContextOrDefault(ctx)
+	awsAccount, err := GetAwsAccountWithIdFromUser(user, id, tx)
+	if err == nil {
+		awsAccount.Pretty = body.Pretty
+		if err := awsAccount.UpdateAWSAccount(ctx, tx); err != nil {
+			return 500, errFailUpdateAccount
```
giubil: You should log an error here

giubil (changes requested):

giubil (dismissed):

aws/aws.go
```diff
@@ -137,6 +137,24 @@ func (a *AwsAccount) CreateAwsAccount(ctx context.Context, db models.XODB) error
 	return err
 }
 
+// UpdateAWSAccount updates an AWS account for a user. It does no error
+// checking: the caller should check themselves that the AWS account exists.
+// Only the Pretty will be updated.
+func (a *AwsAccount) UpdateAWSAccount(ctx context.Context, tx *sql.Tx) error {
```
schubev: If only the `pretty` is updated, the function’s name should reflect it.

aws/aws.go
```diff
@@ -137,6 +137,24 @@ func (a *AwsAccount) CreateAwsAccount(ctx context.Context, db models.XODB) error
...
+	logger := jsonlog.LoggerFromContextOrDefault(ctx)
+	dbAwsAccount, err := models.AwsAccountByID(tx, a.Id)
+	if err != nil {
+		logger.Error("Failed to get AWS account in database.", nil)
+	} else {
+		dbAwsAccount.Pretty = a.Pretty
+		err := dbAwsAccount.Update(tx)
```
schubev: By using `:=` here you shadow the `err` defined at line 145, so at line 155 you would still return `nil`.

schubev: Please write unit tests, you should have caught this with unit tests.

aws/routes.go
```diff
@@ -27,6 +27,13 @@ import (
 	"github.com/trackit/trackit2/users"
 )
 
+var (
+	// QueryArgAwsAccount allows to get the AWS Account ID in the URL Parameters
+	// with routes.RequiredQueryArgs. This AWS Account ID will be an Uint stored in
+	// the routes.Arguments map with itself for key.
+	QueryArgAwsAccount = routes.QueryArg{"aa", "AWS Account ID", routes.QueryArgUint{}}
```
schubev: Isn’t that already defined somewhere else ?

schubev (dismissed):
Beware of variable shadowing.

aws/aws.go
thibautcornolti: `err` is shadowed only if the `err` defined at line 145 is `nil`. It's the expected behaviour.

aws/aws.go
```diff
@@ -137,6 +137,24 @@ func (a *AwsAccount) CreateAwsAccount(ctx context.Context, db models.XODB) error
 	return err
 }
 
+// UpdatePrettyAWSAccount updates an AWS account for a user. It does no error
+// checking: the caller should check themselves that the AWS account exists.
+// Only the Pretty will be updated.
+func (a *AwsAccount) UpdatePrettyAWSAccount(ctx context.Context, tx *sql.Tx) error {
```
schubev: The convention in this project is to treat acronyms as simple words, with only the first letter capitalized, so this should be `UpdatePrettyAwsAccount`.

schubev (changes requested):
Just one change to an identifier, and you’re good !

schubev (approved):

PR #31, opened 2017-12-13
Fix bill repository endpoint

zanchi-r:
This PR fixes the bill repository endpoint by using body parsed by the route decorator instead of trying to decode it again.
Fixes Jira issue TRAC-630

### DIFF
--- a/aws/s3/billRepository.go
+++ b/aws/s3/billRepository.go
@@ -17,7 +17,6 @@ package s3
 import (
 	"context"
 	"database/sql"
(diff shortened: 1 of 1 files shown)


giubil (approved):

giubil (changes requested):
Your commit message should be more explicit. It also does not respect the commit message norm in Go. Please see https://golang.org/doc/contribute.html#commit_changes to see how it should be formatted.

giubil (approved):

PR #33, opened 2017-12-14
Chargify configuration

zanchi-r:
This PR adds a new option to configure the chargify secret and the chargify configuration database model.
This will be used by the closed source part of trackit to implement the chargify integration


schubev:
Code for the closed source “pro” (or “premium” or whatever we’ll call them) features *must not* appear in this repository (or on any public GitHub repository). If code here needs some changes to correctly integrate components from closed source modules, such as configuration flags or database schemas, this is fine as long as those changes are not specific to one module.

PR #37, opened 2018-01-03
costs: changed return value code when ES error

giubil:
TRAC-653 #in-review

### DIFF
--- a/costs/costs_route.go
+++ b/costs/costs_route.go
@@ -178,7 +178,7 @@ func parseFormAndQueryParams(request *http.Request) (esQueryParams, error) {
 	return parsedParams, nil
 }
 
-func makeElasticSearchRequestAndParseIt(ctx context.Context, parsedParams esQueryParams, user users.User) (es.SimplifiedCostsDocument, error) {
+func makeElasticSearchRequestAndParseIt(ctx context.Context, parsedParams esQueryParams, user users.User) (es.SimplifiedCostsDocument, int, error) {
 	l := jsonlog.LoggerFromContextOrDefault(ctx)
 	index := es.IndexNameForUser(user, "lineitems")
 	searchService := GetElasticSearchParams(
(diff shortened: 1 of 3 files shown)


costs/costs_route.go
```diff
@@ -178,7 +178,7 @@ func parseFormAndQueryParams(request *http.Request) (esQueryParams, error) {
 	return parsedParams, nil
 }
 
-func makeElasticSearchRequestAndParseIt(ctx context.Context, parsedParams esQueryParams, user users.User) (es.SimplifiedCostsDocument, error) {
+func makeElasticSearchRequestAndParseIt(ctx context.Context, parsedParams esQueryParams, user users.User) (es.SimplifiedCostsDocument, int, error) {
```
schubev: Because unlike `error` and `es.SimplifiedCostDocument`, `int` does not inform on the purpose of what it contains, perhaps you should document what this function returns.

costs/costs_route.go
```diff
@@ -191,15 +191,20 @@ func makeElasticSearchRequestAndParseIt(ctx context.Context, parsedParams esQuer
 	)
 	res, err := searchService.Do(ctx)
 	if err != nil {
+		if err.Error() == "elastic: Error 404 (Not Found): no such index [type=index_not_found_exception]" {
```
schubev: Instead of comparing the error string, you should use `github.com/olivere/elastic.v5.IsNotFound`.

costs/costs_route.go
```diff
@@ -191,15 +191,20 @@ func makeElasticSearchRequestAndParseIt(ctx context.Context, parsedParams esQuer
...
+			l.Warning("Query execution failed, ES index does not exists : "+index, nil)
```
schubev: It would cost you nothing to log the error as part of the log data. Better to log too much than not enough, IMO.

schubev (changes requested):
Looks good ! Please just add some doc and fix your condition to detect 404s and you’ll be good to go.

schubev (approved):

PR #39, opened 2018-01-09
S3 cost endpoint

zanchi-r:
This adds an endpoint to retrieve details about the cost and usage of S3

TRAC-660

### DIFF
--- a/s3/costs/prepare_response.go
+++ b/s3/costs/prepare_response.go
@@ -51,7 +51,9 @@ func getBucketInfoByName(buckets bucketsInfo, bucketName string) *S3BucketCost {
 	return buckets[bucketName]
 }
 
-func parseBucketsStorage(buckets bucketsInfo, parsedDocument bucket) {
+// parseBucketsStorage parses the cost and usage informations from parsedDocument
+// and insert these informations in the buckets map
+func parseBucketsStorage(buckets bucketsInfo, parsedDocument bucket) bucketsInfo {
 	bucketsField := parsedDocument["buckets"].([]interface{})
 	for _, bucketData := range bucketsField {
 		bucketData := bucketData.(bucket)
@@ -59,22 +61,25 @@ func parseBucketsStorage(buckets bucketsInfo, parsedDocument bucket) {
 		bucketInfo.GbMonth = bucketData["gb"].(bucket)["value"].(float64)
 		bucketInfo.StorageCost = bucketData["cost"].(bucket)["value"].(float64)
 	}
+	return buckets
 }
 
 // parseStorage parses the result from GetS3SpaceElasticSearchParams
-func parseStorage(ctx context.Context, buckets bucketsInfo, resStorage *elastic.SearchResult) error {
+func parseStorage(ctx context.Context, buckets bucketsInfo, resStorage *elastic.SearchResult) (bucketsInfo, error) {
 	var logger = jsonlog.LoggerFromContextOrDefault(ctx)
 	var parsedDocument bucket
 	err := json.Unmarshal(*resStorage.Aggregations["buckets"], &parsedDocument)
 	if err != nil {
 		logger.Error("Failed to parse elasticsearch document.", err.Error())
-		return err
+		return buckets, err
 	}
 	parseBucketsStorage(buckets, parsedDocument)
-	return nil
+	return buckets, nil
 }
 
-func parseBucketsRequests(buckets bucketsInfo, parsedDocument bucket) {
+// parseBucketsRequests parses the cost and usage informations from parsedDocument
+// and insert these informations in the buckets map
+func parseBucketsRequests(buckets bucketsInfo, parsedDocument bucket) bucketsInfo {
 	bucketsField := parsedDocument["buckets"].([]interface{})
 	for _, bucketData := range bucketsField {
 		bucketData := bucketData.(bucket)
@@ -86,22 +91,25 @@ func parseBucketsRequests(buckets bucketsInfo, parsedDocument bucket) {
 			bucketInfo.RequestsCost = bucketData["cost"].(bucket)["value"].(float64)
 		}
 	}
+	return buckets
 }
 
 // parseRequests parses the result from GetS3RequestsElasticSearchParams
-func parseRequests(ctx context.Context, buckets bucketsInfo, resRequests *elastic.SearchResult) error {
+func parseRequests(ctx context.Context, buckets bucketsInfo, resRequests *elastic.SearchResult) (bucketsInfo, error) {
 	var logger = jsonlog.LoggerFromContextOrDefault(ctx)
 	var parsedDocument bucket
 	err := json.Unmarshal(*resRequests.Aggregations["buckets"], &parsedDocument)
 	if err != nil {
 		logger.Error("Failed to parse elasticsearch document.", err.Error())
-		return err
+		return buckets, err
 	}
-	parseBucketsRequests(buckets, parsedDocument)
-	return nil
+	buckets = parseBucketsRequests(buckets, parsedDocument)
+	return buckets, nil
 }
 
-func parseBucketsBandwidth(buckets bucketsInfo, parsedDocument bucket, bwType string) {
+// parseBucketsBandwidth parses the cost and usage informations from parsedDocument
+// and insert these informations in the buckets map
+func parseBucketsBandwidth(buckets bucketsInfo, parsedDocument bucket, bwType string) bucketsInfo {
 	bucketsField := parsedDocument["buckets"].([]interface{})
 	for _, bucketData := range bucketsField {
 		bucketData := bucketData.(bucket)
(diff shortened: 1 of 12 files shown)


s3_costs/es_request_constructor.go
```diff
@@ -0,0 +1,140 @@
...
+//	- durationEnd time.Time : A time.Time struct representing the end of the time range in the query
+//	- client *elastic.Client : an instance of *elastic.Client that represent an Elastic Search client.
+//	It needs to be fully configured and ready to execute a client.Search()
+//	- index string : The Elastic Search index on wich to execute the query. In this context the default value
+//	should be "awsdetailedlineitems"
+// This function excepts arguments passed to it to be sanitize. If they are not, the following cases will make
+// it crash :
+//	- If the client is nil or malconfigured, it will crash
+//	- If the index is not an index present in the ES, it will crash
+func GetS3SpaceElasticSearchParams(accountList []uint, durationBegin time.Time,
```
giubil: The functions `GetS3SpaceElasticSearchParams`, `GetS3RequestsElasticSearchParams` and `GetS3BandwidthElasticSearchParams` have a lot of repeated code into them. Please try merging the repeated code together

s3_costs/s3_costs_route.go
```diff
@@ -0,0 +1,173 @@
...
+	searchService := GetS3RequestsElasticSearchParams(
+		parsedParams.accountList,
+		parsedParams.dateBegin,
+		parsedParams.dateEnd,
+		es.Client,
+		index,
+	)
+	res, err := searchService.Do(ctx)
+	if err != nil {
+		l.Error("Query execution failed : "+err.Error(), nil)
```
giubil: This could be the same insance as what was fixed in the PR #37 where if the index does not exists it will return a 500 instead of returning 200 with no data

s3_costs/s3_costs_route.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, fmt.Errorf("could not execute the ElasticSearch query")
+	}
+	return res, nil
+}
+
+// makeElasticSearchRequestsRequest prepares and run the request to retrieve requests usage/cost
+func makeElasticSearchRequestsRequest(ctx context.Context, parsedParams esQueryParams,
```
giubil: Again, `makeElasticSearchRequestsRequest` and `makeElasticSearchBandwidthRequest` share a lot of the same code. You should merge the redondant code

scripts/add_mapping_es.sh
```diff
@@ -42,6 +42,13 @@ curl \
 				"resourceId": {
 					"type": "keyword"
 				},
+				"usageAmount": {
```
giubil, line 45: I don't think the script is used anymore, it was superseded by the adding of the mapping at startup by @schubev, so there is no need to update it

s3_costs/prepare_response.go
```diff
@@ -0,0 +1,155 @@
...
+// getBucketInfoByName returns the S3BucketCost associated to the bucketName
+func getBucketInfoByName(buckets bucketsInfo, bucketName string) *S3BucketCost {
+	if val, ok := buckets[bucketName]; ok {
+		return val
+	}
+	buckets[bucketName] = &S3BucketCost{}
+	return buckets[bucketName]
+}
+
+func parseBucketsStorage(buckets bucketsInfo, parsedDocument bucket) {
```
giubil: This function does not return anything. Meaning that it will modify arguments passed to it. This is discouraged, as it makes reading the code less clear.
Either change this function to return the changes, or write a comment explaining what is modified

This comment also applies for all instances of functions that does not return modifications

giubil (changes requested):
Pretty good PR, there are just some minor problems such as redondant code, and function that do not return  its modifications.

Also, your package name `s3_costs` does not really respect Go [naming convention](https://golang.org/doc/effective_go.html#mixed-caps). Underscores are not really used in Go, it should either be `s3Costs` or `s3/costs`. The latter would be more appropriate I think, as we could have other things in the `s3` package

s3/costs/s3_costs_route.go
```diff
@@ -0,0 +1,178 @@
...
+
+var (
+	// AwsAccountsQueryArg allows to get the AWS Account IDs in the URL
+	// Parameters with routes.QueryArgs. These AWS Account IDs will be a
+	// slice of Uint stored in the routes.Arguments map with itself for key.
+	AwsAccountsQueryArg = routes.QueryArg{
+		Name:        "accounts",
+		Type:        routes.QueryArgUintSlice{},
+		Description: "The IDs for many AWS account.",
+	}
```
giubil: Here you make the `AwsAccountsQueryArg` not optional, but in the function `GetS3UsageAndCostElasticSearchParams` line 63 of file `s3/costs/es_request_constructor.go` you write a condition that looks like handling for optional `AwsAccounts`

```golang
if len(accountList) > 0 {
 		query = query.Filter(createQueryAccountFilter(accountList))
 	}
```

s3/costs/s3_costs_route.go
```diff
@@ -0,0 +1,178 @@
...
+			routes.QueryArgs{DateEndQueryArg},
+		),
+	}.H().Register("/s3_costs")
+}
+
+var (
+	// AwsAccountsQueryArg allows to get the AWS Account IDs in the URL
+	// Parameters with routes.QueryArgs. These AWS Account IDs will be a
+	// slice of Uint stored in the routes.Arguments map with itself for key.
+	AwsAccountsQueryArg = routes.QueryArg{
```
giubil: Is it really necessary to have all the queryParams exported ?

s3/costs/s3_costs_route.go
```diff
@@ -0,0 +1,178 @@
...
+			users.RequireAuthenticatedUser{},
+			routes.Documentation{
+				Summary:     "get the s3 costs data",
+				Description: "Responds with cost data based on the queryparams passed to it",
+			},
+			routes.QueryArgs{AwsAccountsQueryArg},
+			routes.QueryArgs{DateBeginQueryArg},
+			routes.QueryArgs{DateEndQueryArg},
+		),
+	}.H().Register("/s3_costs")
```
giubil: You modified the package name from `s3_costs` to `s3/costs`. Wouldn't it be better to have the route url reflect that as well ?

giubil (dismissed):
Just some minor details to fix up, and it should be good to merge

s3/costs/prepare_response.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		logger.Error("Failed to parse elasticsearch document.", err.Error())
+		return buckets, err
+	}
+	buckets = parseBuckets(buckets, parsedDocument, resultType)
+	return buckets, nil
+}
+
+// prepareResponse parses the results from elasticsearch and returns a map of buckets with their usage informations
+func prepareResponse(ctx context.Context, resStorage, resRequests, resBandwidthIn, resBandwidthOut *elastic.SearchResult) (interface{}, error) {
```
schubev, line 130: Don’t repeat yourself! What about something like this?
```go
func prepareResponse(
    ctx context.Context,
    resStorage, resRequests, resBandwidthIn, resBandwidthOut *elastic.SearchResult
) (interface{}, error) {
    var buckets = make(bucketsInfo)
    var err error
    var components = [...]struct{k string, sr *elastic.SearchResult}{
        { "storage", resStorage },
        { "requests", resRequests },
        { "bandwidthIn", resBandwidthIn },
        { "bandwidthOut", resBandwidthOut },
    }
    for _, cpn := range components {
        buckets, err = parseESResult(ctx, buckets, cpn.sr, cpn.k)
        if err != nil {
           return nil, err
        }
    }
    return buckets, nil
}
```

s3/costs/s3_costs_route.go
```diff
@@ -0,0 +1,178 @@
...
+			return nil, http.StatusOK, err
+		}
+		l.Error("Query execution failed : "+err.Error(), nil)
+		return nil, http.StatusInternalServerError, fmt.Errorf("could not execute the ElasticSearch query")
+	}
+	return res, http.StatusOK, nil
+}
+
+// getS3CostData returns the s3 cost data based on the query params, in JSON format.
+func getS3CostData(request *http.Request, a routes.Arguments) (int, interface{}) {
```
schubev, line 151: Same here, please don’t repeat yourself.

s3/costs/s3_costs_route.go
```diff
@@ -0,0 +1,178 @@
...
+		),
+	}.H().Register("/s3_costs")
+}
+
+var (
+	// AwsAccountsQueryArg allows to get the AWS Account IDs in the URL
+	// Parameters with routes.QueryArgs. These AWS Account IDs will be a
+	// slice of Uint stored in the routes.Arguments map with itself for key.
+	AwsAccountsQueryArg = routes.QueryArg{
+		Name:        "accounts",
```
schubev: Elsewhere in the project, we used `aas` as the name for this parameter, please change to this so as to have a consistent API.

schubev (changes requested):

s3/costs/prepare_response.go
zanchi-r: Done!

s3/costs/s3_costs_route.go
zanchi-r: Done!

s3/costs/s3_costs_route.go
zanchi-r: Done

schubev (approved):

PR #40, opened 2018-01-10
Use QueryArgs for /costs route

giubil:
Added `QueryArgStringSlice` and `QueryArgDate` in the `routes` package, and changed the `/cost` route to use those

### DIFF
--- a/costs/costs_route.go
+++ b/costs/costs_route.go
@@ -18,7 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"net/url"
+	"strconv"
 	"strings"
 	"time"
 
@@ -29,10 +29,6 @@ import (
 	"github.com/trackit/trackit2/users"
 
 	"gopkg.in/olivere/elastic.v5"
-)
-
-const (
-	iso8601TimeFormat = "2006-01-02"
 )
 
 // simpleCriterionMap will map simple criterion to the boolean true.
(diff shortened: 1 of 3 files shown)


costs/costs_route.go
```diff
@@ -56,128 +52,75 @@ type esQueryParams struct {
...
-	}
-	parsedParams, err := parseQueryParams(request.Form)
-	if err != nil {
-		l.Info("Error parsing user submitted forms : "+err.Error(), nil)
-		return esQueryParams{}, err
+// validateAwsAccounts will validate awsAccounts passed to it.
+// It checks that they are numbers that are 12 character long
+func validateAwsAccounts(parsedParams esQueryParams) error {
+	for _, account := range parsedParams.accountList {
+		if _, err := strconv.ParseInt(account, 10, 0); err != nil || len(account) != 12 {
```
schubev, line 119: Can you please point me to the documentation which says an AWS account ID is always 12 digits long ?

costs/costs_route.go
```diff
@@ -56,128 +52,75 @@ type esQueryParams struct {
...
+			return fmt.Errorf("Error invalid account format : %s", account)
```
schubev: Please use the Go convention for errors.

schubev (changes requested):

costs/costs_route.go
giubil: It's written here : https://docs.aws.amazon.com/general/latest/gr/acct-identifiers.html

schubev: Thanks!

schubev (approved):

PR #41, opened 2018-01-15
aws: added the DELETE /aws route to delete an AwsAccount

giubil:
TRAC-576

### DIFF
--- a/aws/routeAwsDelete.go
+++ b/aws/routeAwsDelete.go
@@ -53,11 +53,11 @@ func deleteAwsAccount(r *http.Request, a routes.Arguments) (int, interface{}) {
 	if err != nil {
 		l.Error("error while deleting user's AWS account", err.Error())
 		if res == -1 {
-			l.Error("Failed to retrieve user's AWS accounts.", err.Error())
-			return http.StatusInternalServerError, errors.New("Failed to retrieve user's AWS accounts")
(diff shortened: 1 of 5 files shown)


schubev (changes requested):
You don’t need to retrieve the user’s AWS accounts to delete one. When you delete a row from the database, `db.Exec` gives you the number of deleted rows, so you can just do `DELETE FROM trackit.aws_account WHERE id = ? AND userId = ?` and check that at least one row was deleted.

aws/routeAwsDelete.go
```diff
@@ -0,0 +1,64 @@
...
+func deleteAwsAccount(r *http.Request, a routes.Arguments) (int, interface{}) {
+	u := a[users.AuthenticatedUser].(users.User)
+	tx := a[db.Transaction].(*sql.Tx)
+	l := jsonlog.LoggerFromContextOrDefault(r.Context())
+	accountToDeleteID := a[AwsAccountQueryArg].(int)
+	res, err := DeleteAwsAccountFromAccountID(tx, u.Id, accountToDeleteID)
+	if err != nil {
+		l.Error("error while deleting user's AWS account", err.Error())
+		if res == -1 {
+			l.Error("Failed to retrieve user's AWS accounts.", err.Error())
```
giubil: When logging an error, you shouldn't either capitalise nor finish the sentence by a period

schubev (approved):

PR #43, opened 2018-01-16
Improved README

zanchi-r:
This PR improves the look of the README and adds more detailed informations on how to run TrackIt

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,18 +1,62 @@
+[#trackit-logo]
+image::https://s3-us-west-2.amazonaws.com/trackit-public-artifacts/github-page/logo.png[]
+
 = Trackit (v2)
 
 This is the new iteration of MSolution’s Trackit, a tool to optimize your AWS
 usage and spending.
 
-== Cheking out
+== Don't want to self-host trackit?
 
-In order to check out this project’s dependencies, run `govendor sync -v`.
+We run our own version, check it out at https://trackit.io/[TrackIt.IO]
 
-== Running
+= Installation
 
-In order to build and run the _Docker Compose_ environment, simply run
-`./start.sh`.
+=== #0 Be sure all requirements below are met
 
-== API documentation
+- Install https://docs.docker.com/engine/installation/[Docker]
+- Install https://docs.docker.com/compose/install/[Docker Compose]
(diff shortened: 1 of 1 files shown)


lfrancois (commented):
@zanchi-r: very good start, but I feel the README, should be in the https://github.com/trackit/trackit repo and describe how to install the server and the webui.

The goal is to point people to https://github.com/trackit/trackit and allow them to deploy without to even have to think about where is the server or the ui

zanchi-r:
I think each repository should have a README that explains how to deploy the content of the repository.
The trackit repository will have a README that explains how to deploy the full solution, but I can't write it now since we have not defined the deployment process yet

README.adoc
```diff
@@ -1,18 +1,62 @@
+[#trackit-logo]
+image::https://s3-us-west-2.amazonaws.com/trackit-public-artifacts/github-page/logo.png[]
```
schubev, line 2: I think the logo should be smaller (at least 50%) ; this is a bit burlesque.

README.adoc
```diff
@@ -1,18 +1,62 @@
...
+----
+$> go get -u github.com/kardianos/govendor
+$> export PATH=$PATH:$HOME/go/bin
+----
+
+=== #1 Clone TrackIt
+
+[source,sh]
+----
+$> mkdir -p $HOME/go/src/github.com/trackit && cd $HOME/go/src/github.com/trackit
```
schubev, line 33: This is not portable, you should use `$GOPATH` instead.

schubev (changes requested):

README.adoc
zanchi-r: I used $HOME/go to match how they do it on the [Go Getting started](http://github.com) documentation.
I think users who don't know go will not set the GOPATH since it's not mandatory and they may have issues running the server, whereas people that are familiar with go will understand that they can put it in the GOPATH.
But if you think it's a bad idea I can change it!

lfrancois: You need to remember, we need to make it as easy as possible for people to test it. Following the official go documentation to do it sounds like a good principle.

README.adoc
lfrancois: It's a detail for now. We can change it later on.

lfrancois (approved):
LGTM

README.adoc
schubev: Then perhaps the best solution would be to simply provide Docker images for users, and assume people who may want to modify the code know Go and how to manage their `$GOPATH`.

PR #44, opened 2018-01-23
Prevent data race condition in the CSV decoder

zanchi-r:
This prevents a data race condition when multiple CSV decoders try to access the shared recordTypeCache variable at the same time

### DIFF
--- a/util/csv/csv.go
+++ b/util/csv/csv.go
@@ -5,6 +5,12 @@ import (
 	"io"
 	"reflect"
 	"strings"
+	"sync"
+)
+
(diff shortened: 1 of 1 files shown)


schubev (changes requested):
So you found the issue, great !

AFAIK Go maps are safe to read concurrently, so you can reduce the impact on performance by using a `sync.RWMutex` instead of `sync.Mutex` so as to allow concurrent reads.

Please change that and this PR will be good to go.

schubev (approved):

PR #45, opened 2018-01-24
Prevent sigsegv in case of elasticsearch error during bill ingestion

zanchi-r:
In case of elasticsearch error the BulkResponse is nil, therefore trying to access resp.Took creates a SIGSEGV


schubev:
Can you try and understand why `resp` would be nil ?

zanchi-r:
It seems to be the expected behaviour:

The return value is set here: https://github.com/olivere/elastic/blob/v5.0.62/bulk_processor.go#L474

And BulkService.Do returns nil in case of error when performing the request:
https://github.com/olivere/elastic/blob/v5.0.62/bulk.go#L254

schubev (approved):

PR #47, opened 2018-02-19
aws/s3: fix ES mapping

schubev:
Fix the ElasticSearch mapping template so that lineitems are stored with correct types.


zanchi-r (approved):
LGTM

PR #48, opened 2018-02-20
Billing diff

zanchi-r:
This PR adds two endpoints:

- The absolute endpoint that returns the absolute usagetype cost for each week/month in the time range
- The variations endpoint that returns the usagetype cost variations between each pair of week/month in the time range

Both these endpoints can return JSON or CSV depending on the content type specified by the user


routes/routes.go
```diff
@@ -51,20 +53,38 @@ type Decorator interface {
...
 	status, output := h.Func(w, r, arguments)
-	w.Header()["Content-Type"] = []string{"application/json; charset=utf-8"}
-	w.WriteHeader(status)
-	e := json.NewEncoder(w)
-	if config.PrettyJsonResponses {
-		e.SetIndent("", "\t")
+	contentType := r.Header["Content-Type"][0]
+	w.Header()["Content-Type"] = []string{fmt.Sprintf("%s; charset=utf-8", contentType)}
+	switch contentType {
+	case "application/json":
```
schubev: HTTP Content-Type negociation is more complicated than that : user agents can specify a list of accepted content types, with prioritization, and usage of wildcards. For example a client may request ```Accept: text/*;q=0.5 application/json;q=0.8``` where `application/json` should be selected. I think this should be implement as a multiplexer like `routes.MethodMuxer` and as a decorator.

costs/diff/diff_route.go
```diff
@@ -0,0 +1,207 @@
...
+type esQueryParams struct {
+	dateBegin         time.Time
+	dateEnd           time.Time
+	accountList       []string
+	aggregationPeriod string
+}
+
+// diffQueryArgs allows to get required queryArgs params
+var diffQueryArgs = []routes.QueryArg{
+	routes.QueryArg{
```
schubev: You should reuse `aws.AwsAccountsOptionalQueryArgs` instead of redefining it.

costs/diff/diff_route.go
```diff
@@ -0,0 +1,207 @@
...
+		Name:        "aas",
+		Description: "List of comma separated AWS accounts ids",
+		Type:        routes.QueryArgStringSlice{},
+		Optional:    true,
+	},
+	routes.QueryArg{
```
schubev: You should try and reuse `QueryArg` instances.

costs/diff/diff_route.go
```diff
@@ -0,0 +1,207 @@
...
+				Summary:     "get the cost diff variations",
+				Description: "Responds with the cost diff variations based on the query args passed to it",
+			},
+		),
+	}.H().Register("/costs/diff/variations")
+}
+
+// validateAwsAccounts will validate awsAccounts passed to it.
+// It checks that they are numbers that are 12 character long
+func validateAwsAccounts(parsedParams esQueryParams) error {
```
schubev: Doesn’t this function already exist in the cost breakdown implementation ?

schubev (dismissed):

giubil (approved):

PR #53, opened 2018-03-28
Reports endpoints

zanchi-r:
This PR adds two endpoints:
- An endpoint to retrieve the list of reports available for the current user
- An endpoint to download a report

### DIFF
--- a/reports/reports_route.go
+++ b/reports/reports_route.go
@@ -1,0 +1,141 @@
+//   Copyright 2017 MSolution.IO
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


reports/reports_route.go
```diff
@@ -0,0 +1,141 @@
...
+	downloader := s3manager.NewDownloader(awsSession.Session)
+	_, err := downloader.Download(buff,
+		&s3.GetObjectInput{
+			Bucket: aws.String(config.ReportsBucket),
+			Key:    aws.String(reportPath),
+		})
+	if err != nil {
+		return http.StatusNotFound, fmt.Errorf("The specified key does not exist")
+	}
+	return http.StatusOK, Report{buff.Bytes(), reportName}
```
thibautcornolti, line 140: Does that buffer the file? The report could be really huge.

zanchi-r: @thibautcornolti Yes it does but I'm not sure there's a better way to do it. For now the biggest report I generated was 83kb so not really an issue

thibautcornolti (approved):

reports/reports_route.go
thibautcornolti: If it does, it's good to me.

PR #54, opened 2018-03-29
fixed db schema to match migrations

zanchi-r:


thibautcornolti (approved):
Considering that works.

PR #57, opened 2018-06-25
Add support for basic teams.

schubev:
Adds basic support for teams. By “basic” I mean that teams are actually a normal user, and a set of “viewer users”. Viewer users connect just like normal users ( they are actually users, stored in the `user` table, however they have a “parent user” ), and should be able to see anything the parent user can see. Each route which requires authentication can treat viewer users as themselves or as their parent, or hit them with a 403 ( this is the default so that if we ever forget a route, it won’t be a security issue. See RequireAuthenticatedUser ).
When they are created the server generates a password for them. This password is hashed and the cleartext is never stored, however it is sent in the response to the client, so that they can provide it to their coworker/teammate/whatever. The viewer user can then change their password.

To have real teams ( multiple teams per user, roles within teams, invitations ) we need to implement email workflows.


thibautcornolti (approved):
Really smart way to handle the viewer users, even if it makes the viewer user unable to be also a classic user. I guess we'll never need to do that.

However, we definitely need an email workflow, at least to check the viewer user mail and to hide if this account already exists.

All seem really good and clear to me, and you've respected the coding style. I approve.

hug33k (approved):

PR #63, opened 2018-07-17
Change bill status for bill error

simonmeyerrr:
I changed every bill status into bill error (Hug33k's request) and i modified the column in the DB into NOT NULL DEFAULT "" because NULL created problem with Go.


thibautcornolti (approved):
Ok!

PR #64, opened 2018-07-18
Feature/error message existing account

simonmeyerrr:
This feature add a specific error message in the response of the API when you sign up and the account already exists.


users/create.go
```diff
@@ -118,7 +119,12 @@ func createUserWithValidBody(request *http.Request, body createUserRequestBody,
 		return 200, user
 	} else {
 		logger.Error(err.Error(), nil)
-		return 500, errors.New("Failed to create user.")
+		errSplit := strings.Split(err.Error(), ":")
+		if (len(errSplit) >= 1 && errSplit[0] == "Error 1062") {
+			return 409, errors.New("Account already exists.")
```
thibautcornolti: Why 409 ? I think 400 is better.

thibautcornolti:
You could add the same enhancement with the _viewer users_. (line 89 I guess)

thibautcornolti (approved):

PR #74, opened 2018-07-20
Feature/verify bill validty on submit

simonmeyerrr:
API returns now an error message and a 400 status when the bill location isn't accessible on Post and Patch method.

### DIFF
--- a/aws/s3/billRepository.go
+++ b/aws/s3/billRepository.go
@@ -238,8 +238,6 @@ func postBillRepository(r *http.Request, a routes.Arguments) (int, interface{}) 
 	return postBillRepositoryWithValidBody(r, tx, aa, body)
 }
 
-
-
 func postBillRepositoryWithValidBody(
 	r *http.Request,
 	tx *sql.Tx,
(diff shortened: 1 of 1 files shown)


aws/s3/billRepository.go
```diff
@@ -344,6 +345,17 @@ func isPrefixValid(p string) error {
 	}
 }
 
+func isBillRepositoryAccessible(aa aws.AwsAccount, body postBillRepositoryBody) (error) {
+	ctx := context.Background()
```
thibautcornolti: Don't use the Background context, you can have one by doing `r.Context()`.

aws/s3/billRepository.go
```diff
@@ -344,6 +345,17 @@ func isPrefixValid(p string) error {
...
+	l := jsonlog.LoggerFromContextOrDefault(ctx)
+	_, _, err := getServiceForRepository(ctx, aa, BillRepository{Bucket: body.Bucket, Prefix: body.Prefix})
+	if (err != nil) {
+		l.Warning("Trying to add a bad bill location.", err.Error())
+		return errors.New("Can't acces to this bill location.")
```
thibautcornolti: Access

thibautcornolti: `Couldn't` is better

thibautcornolti (changes requested):

thibautcornolti (approved):
Good!

PR #75, opened 2018-07-25
Current month report

simonmeyerrr:
Remove the condition that blocks the ingestion of incomplete bills. Added a function that removes incomplete bills in the elastic search of a specific bill repository before ingesting new invoices.


es/billingRepository.go
```diff
@@ -27,3 +27,11 @@ func CleanByBillRepositoryId(ctx context.Context, aaUId, brId int) error {
 	_, err := elastic.NewDeleteByQueryService(Client).Index(index).Query(query).Do(ctx)
 	return err
 }
+
+func CleanCurrentMonthBillByBillRepositoryId(ctx context.Context, aaUId, brId int) error {
```
thibautcornolti: Please add comments.

aws/s3/ingestBills.go
```diff
@@ -70,6 +70,7 @@ func reportUpdateConclusionChanToSlice(rucc <-chan ReportUpdateConclusion, count
 // them.
```
thibautcornolti: Update the comment.

aws/s3/ingestBills.go
```diff
@@ -89,6 +90,10 @@ func UpdateDueReports(ctx context.Context, tx *sql.Tx) ([]ReportUpdateConclusion
 			}
 			aas[br.AwsAccountId] = aa
 		}
+		err = es.CleanCurrentMonthBillByBillRepositoryId(ctx, aa.UserId, br.Id)
+		if err != nil {
+			l.Info("No incomplete bills found for this bill repository.", err)
```
thibautcornolti: Does ES return an error if there's nothing to delete ? If it doesn't, `Error when cleaning the current month.` is more appropriate as well as Error instead of Info.

Please add the account `aa` and the bill repository `br` to the log.

es/billingRepository.go
```diff
@@ -27,3 +27,11 @@ func CleanByBillRepositoryId(ctx context.Context, aaUId, brId int) error {
...
+	index := IndexNameForUserId(aaUId, IndexPrefixLineItems)
+	query := elastic.NewBoolQuery()
+	query = query.Filter(elastic.NewTermQuery("billRepositoryId", brId), elastic.NewTermQuery("invoiceId", ""))
+	_, err := elastic.NewDeleteByQueryService(Client).Index(index).Query(query).Do(ctx)
+	return err
+}
```
thibautcornolti: New line

thibautcornolti (changes requested):

aws/s3/ingestBills.go
simonmeyerrr: It return an error when there aren't any element to delete.

thibautcornolti: Maybe you could log an error if there's one, except if it's a _Not found_, just skip the log.

es/billingRepository.go
```diff
@@ -28,10 +28,11 @@ func CleanByBillRepositoryId(ctx context.Context, aaUId, brId int) error {
 	return err
 }
 
+// Remove incomplete bills of a specific bill repository (invoiceId == "" when incomplete)
```
thibautcornolti: Comments always starts with the function name.

aws/s3/ingestBills.go
```diff
@@ -66,7 +66,7 @@ func reportUpdateConclusionChanToSlice(rucc <-chan ReportUpdateConclusion, count
 	return
 }
 
-// UpdateDueReports finds all BillRepositories in need of an update and updates
+// UpdateDueReports finds all BillRepositories in need of an update, delete old incomplete reports and updates
```
thibautcornolti: delete**s**

thibautcornolti (changes requested):

aws/s3/ingestBills.go
```diff
@@ -89,6 +90,14 @@ func UpdateDueReports(ctx context.Context, tx *sql.Tx) ([]ReportUpdateConclusion
 			}
 			aas[br.AwsAccountId] = aa
 		}
+		err = es.CleanCurrentMonthBillByBillRepositoryId(ctx, aa.UserId, br.Id)
```
thibautcornolti, line 93: You've cleaned when we call ingest-due task but you've forgotten the ingest task. Ingest task is the only one used in production.

thibautcornolti (changes requested):

PR #80, opened 2018-08-02
Current month report

simonmeyerrr:
This update allows the data of the current month to be ingested.
Operation:
- Get the modified manifests at least 1 month before the last ingestion
- delete incomplete data from the BillRepository concerned in the elastic search
- select in the manifests those which are incomplete or have been modified after the last import
- put the manifests in the elastic search


aws/s3/readBills.go
```diff
@@ -408,7 +406,7 @@ func listBillsFromRepositoryPage(
 	count := 0
 	return func(page *s3.ListObjectsV2Output, last bool) bool {
 		for _, o := range page.Contents {
-			if brr.LastImportedManifest.Before(*o.LastModified) {
+			if brr.LastImportedManifest.Before((*o.LastModified).Add(time.Hour * 24 * 30)) {
```
zanchi-r: If you want to add a month, wouldn't it be better to use AddDate(years int, months int, days int)?

es/billingRepository.go
```diff
@@ -20,10 +20,20 @@ import (
...
 	_, err := elastic.NewDeleteByQueryService(Client).WaitForCompletion(false).Index(index).Query(query).Do(ctx)
 	return err
 }
+
+// CleanCurrentMonthBillByBillRepositoryId removes incomplete bills of a specific bill repository (invoiceId == "" when incomplete)
+func CleanCurrentMonthBillByBillRepositoryId(ctx context.Context, aaUId, brId int) error {
+	index := IndexNameForUserId(aaUId, IndexPrefixLineItems)
+	query := elastic.NewBoolQuery()
+	query = query.Filter(elastic.NewTermQuery("billRepositoryId", brId), elastic.NewTermQuery("invoiceId", ""))
+	_, err := elastic.NewDeleteByQueryService(Client).Index(index).Query(query).Do(ctx)
```
zanchi-r: You should not wait for completion when starting the query (see CleanBillRepositoryId), it can cause timeouts when there is a lot of data to clean

aws/s3/ingestBills.go
```diff
@@ -181,11 +181,19 @@ func ingestLineItems(ctx context.Context, bp *elastic.BulkProcessor, index strin
 // manifestsStartingAfter returns a manifest predicate which is true for all
 // manifests starting after a given date.
 func manifestsModifiedAfter(t time.Time) ManifestPredicate {
-	return func(m manifest) bool {
-		if time.Time(m.LastModified).After(t) {
-			return true
+	return func(m manifest, oneMonthBefore bool) bool {
+		if (oneMonthBefore) {
+			if time.Time(m.LastModified).Add(time.Hour * 24 * 30).After(t) {
```
zanchi-r: If you want to add a month, wouldn't it be better to use AddDate(years int, months int, days int)?

zanchi-r (changes requested):
Just a few details and it should be ready to merge.
Please base the PR on stg instead of master.

zanchi-r (approved):

PR #96, opened 2018-08-23
Add IAM policy info

simonmeyerrr:
I added a policies folder that works on the same principle as DB:

 - a file _all_policies.json_ which contains all the necessary policies for the current trackit version, you will need to update it when you add new calls to AWS.
 - a _tool_policies_ folder that contains the policies for each feature to update as well

Although I added a policy for monitoring by analyzing files, I'm not sure about it because I couldn't test it.


policies/tool_policies/monitor_ec2.json
```diff
@@ -0,0 +1,16 @@
+{
```
zanchi-r: This file should be called monitor_resources.json or something similar since it includes other resources than EC2

zanchi-r (changes requested):

zanchi-r (approved):

PR #109, opened 2018-09-04
Fix : Accept token from aws

Lolincolc:
This PR is a quick fix that could fix the issue with AWS Integration and "Error when checking the AWS token" error message.

### DIFF
--- a/users/create.go
+++ b/users/create.go
@@ -35,6 +35,7 @@ import (
 	"github.com/trackit/trackit-server/mail"
 	"github.com/satori/go.uuid"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


users/create.go
```diff
@@ -27,14 +27,14 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/trackit/jsonlog"
+	"github.com/satori/go.uuid"
+	"github.com/aws/aws-sdk-go/aws/awserr"
```
simonmeyerrr: put it next to the other aws import

users/create.go
```diff
@@ -116,9 +116,13 @@ func checkAwsTokenLegitimacy(ctx context.Context, token string) (*marketplacemet
 	svc := marketplacemetering.New(mySession)
 	awsInput.SetRegistrationToken(token)
 	result, err := svc.ResolveCustomer(&awsInput)
+	aerr, ok := err.(awserr.Error)
```
simonmeyerrr: do it differently:
```
if err != nil {
  [cast err into aws]
  if ok {
     [display aws err]
  } else {
   [display error normal]
  }
}
```

simonmeyerrr (changes requested):

simonmeyerrr (approved):

PR #129, opened 2018-10-01
Plugins task

zanchi-r:
This PR adds:
- a new background task to run account plugins
- Some utilities functions and structures that are used to create plugins
- the unused EBS plugin

### DIFF
--- a/plugins/account/core/core.go
+++ b/plugins/account/core/core.go
@@ -1,0 +1,72 @@
+//   Copyright 2018 MSolution.IO
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


plugins/account/core/core.go
```diff
@@ -0,0 +1,72 @@
...
+type PluginParams struct {
+	Context            context.Context
+	User               users.User
+	AwsAccount         aws.AwsAccount
+	AccountId          string
+	AccountCredentials *credentials.Credentials
+	ESClient           *elastic.Client
+}
+
+// PluginParams is the struct that each plugin should return
```
simonmeyerrr: bad struct name

plugins/account/unusedEBS/unusedEBS.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+
+	core "github.com/trackit/trackit-server/plugins/account/core"
+	"github.com/trackit/trackit-server/plugins/utils"
+)
+
+func init() {
+	// Register the plugin
```
simonmeyerrr, line 29: Comment should be out the function

simonmeyerrr (approved):
It works

PR #143, opened 2018-10-12
Pending status for new accounts

Lolincolc:
This PR change pending status for new accounts. New accounts will be marked as "pending" when created and until first ingestion completion.


zanchi-r (changes requested):
Some ingestions are too big to be done at once on account creation. 
For example the trackit account needs two ingestions, with your method it would be returned as completed after the first ingestion even if the last imported manifest was in may.
I think a best way to do it would be to check if the last manifest is less than a week, or something like that

simonmeyerrr (approved):

zanchi-r (approved):

PR #154, opened 2018-10-24
Unattached EIP plugin

zanchi-r:
This PR adds the unattached EIP plugin


thibautcornolti (dismissed):
Go routines would be really more efficient.

thibautcornolti (approved):

PR #157, opened 2018-10-26
S3 traffic plugin

zanchi-r:
This PR adds the s3 traffic plugin along with a new configuration parameter to create payer account plugins

### DIFF
--- a/plugins/README.adoc
+++ b/plugins/README.adoc
@@ -25,11 +25,12 @@ There are a few rules a plugin must respect in order to work:
 ----
 func init() {
 	core.AccountPlugin{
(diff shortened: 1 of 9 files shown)


thibautcornolti (commented):
`PayerAccountOnly: false, // Set to true if billings are required`
Payer accounts have billings but doesn't spend anything. If you run the plugin on a payer account, it will show almost nothing.

Your ES requests don't have any account filter, I'm not sure to understand the point of this plugin. It will mix every data of every AWS account on the TrackIt account right?

Could you add a better description to the PR?

thibautcornolti (approved):

PR #161, opened 2018-11-05
[TRAC-1046] AWS sub accounts

simonmeyerrr:
This PR adds:

### - A new SQL table
This table stores sub accounts linked to an aws account.

### - A function that update sub accounts
This function is called when you Post a new aws account and when a task ingest is finished.

### - An array of sub accounts for each account on the endpoint /aws?detailled


zanchi-r:
Looks good for now, but we also need to update the subaccounts when doing a PATCH on an account

simonmeyerrr:
> Looks good for now, but we also need to update the subaccounts when doing a PATCH on an account

I checked: PATCH only changes the name or the boolean `Payer`. It doesn't change the RoleArn or the External

PR #167, opened 2018-11-09
Endpoint to get AWS Accounts status

hug33k:
This PR adds a new endpoint (`/aws/status`) that returns every users' AWS Account with their bill repositories and also their statuses.

This endpoint cas be used to check if there is an issue with AWS Account or bill repositories configuration.


aws/routes/routeAwsStatus.go
```diff
@@ -0,0 +1,153 @@
...
+		)
+		res[id] = item
+		if err != nil {
+			return nil, err
+		}
+	}
+	return res, nil
+}
+
+func getAwsAccountsStatus(r *http.Request, a routes.Arguments) (int, interface{}) {
```
thibautcornolti: You should keep your types. `(int, []awsAccountWithStatus, error)` would be better.

aws/routes/routeAwsStatus.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	} else {
+		return status{
+			Value: "ok",
+			Detail: "",
+		}
+	}
+}
+
+func getLastJobs(tx *sql.Tx, billRepositoriesIds []int) (map[int]job, error) {
```
thibautcornolti: Any heavy SQL code should be located in `models/awsbillupdatejob.go`.

aws/routes/routeAwsStatus.go
```diff
@@ -0,0 +1,153 @@
...
+	var sqlstr = `
```
thibautcornolti: You should create a view and, if you can, get values with XO.

thibautcornolti (changes requested):

aws/routes/routeAwsStatus.go
```diff
@@ -0,0 +1,153 @@
...
+		for _, billRepository := range awsAccount.BillRepositories {
+			billRepositoriesIds = append(billRepositoriesIds, billRepository.Id)
+		}
+	}
+	jobs, err := getLastJobs(tx, billRepositoriesIds)
+	if err != nil {
+		l.Error("failed to get AWS accounts' bill repositories import statuses", err.Error())
+		return 500, errors.New("failed to retrieve import statuses")
+	}
+	result := make([]interface{}, 0)
```
thibautcornolti: No need to use `interface` here.

zanchi-r (approved):

PR #168, opened 2018-11-09
List of sub accounts

simonmeyerrr:


aws/aws.go
```diff
@@ -177,12 +186,38 @@ func (a *AwsAccount) UpdatePrettyAwsAccount(ctx context.Context, tx *sql.Tx) err
...
 // GetAwsAccountIdentity returns the AWS identity of an AWS Account.
 func (a *AwsAccount) GetAwsAccountIdentity() (identity string, err error) {
-	if reg, err := regexp.Compile("^arn:aws:iam::([0-9]{12}):.*$"); err != nil {
+	if a.RoleArn == "" {
+		return a.AwsIdentity, nil
+	} else if reg, err := regexp.Compile("^arn:aws:iam::([0-9]{12}):.*$"); err != nil {
 		return "", err
 	} else {
 		identity = reg.FindStringSubmatch(a.RoleArn)[1]
+		a.AwsIdentity = identity
```
thibautcornolti: `Get` should not set.

aws/sub_account.go
```diff
@@ -0,0 +1,81 @@
...
+	creds, err := GetTemporaryCredentials(aa, "GetAwsSubAccounts")
+	if err != nil {
+		return nil, err
+	}
+	sess := session.Must(session.NewSession(&aws.Config{
+		Credentials: creds,
+		Region:      aws.String(config.AwsRegion),
+	}))
+	orga := organizations.New(sess)
+	res, err := orga.ListAccounts(nil)
```
thibautcornolti, line 60: Did you update the policy file?

aws/sub_account.go
```diff
@@ -0,0 +1,81 @@
...
+		for _, old := range alreadyAccounts {
+			if old.AwsIdentity == sub.AwsIdentity {
+				already = true
+				break
+			}
+		}
+		if !already {
+			sub.CreateAwsAccount(ctx, tx)
+		}
+	}
```
thibautcornolti:
```
SubAccountLoop:
	for _, sub := range subAccounts {
		for _, old := range alreadyAccounts {
			if old.AwsIdentity == sub.AwsIdentity {
				continue SubAccountLoop
			}
		}
		sub.CreateAwsAccount(ctx, tx)
	}
```

server/taskIngest.go
```diff
@@ -88,10 +88,46 @@ func ingestBillingDataForBillRepository(ctx context.Context, aaId, brId int) (er
 			"error":            err.Error(),
 		})
 	}
+	updateIdentityAndSubAccounts(ctx, aa)
```
thibautcornolti: Note for the future: We should add another task to check sub accounts. This task would add missing sub accounts, remove obsolete ones and perhaps update their AwsIdentity field.
What do you think ?

thibautcornolti: Shoudn't we updateIdentity when creating or editing an AWS Account as well?

thibautcornolti (changes requested):

server/taskIngest.go
simonmeyerrr: It is done when creating, edit can only change pretty and payer so we don't need to check at this moment. And this is done here to add identity for old aws accounts

simonmeyerrr: Yes, I wanted to do it on a separated task but for now, it's more easy for romain to keep it in this task.

thibautcornolti (approved):

zanchi-r (approved):

PR #172, opened 2018-11-16
added endpoint to link a subaccount

zanchi-r:
This PR adds the PATCH /aws/subaccount endpoint to add a role/external to a subaccount


thibautcornolti (approved):

thibautcornolti (dismissed):
We should add a check if the account is actually a sub account

simonmeyerrr (approved):

PR #180, opened 2018-11-23
Spreadsheet reports improvement

hug33k:
This PR adds some improvements to the spreadsheet reports generation :

- Spreadsheet reports generation is no longer executed with task `process-account` (Before, it was executed each time new monthly reports were created).
- Task `generate-spreadsheet` will only create a spreadsheet if some conditions are filled
- Styling is now available for Spreadsheets (Text formatting, color, cell value type, ...)


reports/spreadsheet_styling.go
```diff
@@ -0,0 +1,79 @@
+package reports
```
simonmeyerrr, line 15: missing header

reports/rds.go
```diff
@@ -18,57 +18,61 @@ import (
 	"context"
 	"database/sql"
 	"github.com/trackit/jsonlog"
```
simonmeyerrr, line 21: missing `\n`

simonmeyerrr (changes requested):
Compilation don't work: you forget to update the argument in `server/taskProcessAccount` l.76

simonmeyerrr (approved):

PR #198, opened 2018-12-13
Task to compare es cost vs explorer cost

zanchi-r:
This PR adds a task that we will use to check if the costs are correct on the trackit account.
It compares the costs from ES and from the explorer over the last 6 months.

It does not require an in depth review, the task will be only run for the trackit account and is for debug purpose. It does not affect any feature

### DIFF
--- a/server/taskCheckCost.go
+++ b/server/taskCheckCost.go
@@ -1,0 +1,168 @@
+//   Copyright 2018 MSolution.IO
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


server/taskCheckCost.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	if len(explorerCosts.ResultsByTime) != len(esCosts.Children) {
+		logger.Error("Number of month does not match for cost comparison", map[string]interface{}{
+			"explorerMonthNb": len(explorerCosts.ResultsByTime),
+			"esMonthNb":       len(esCosts.Children),
+		})
+		return
+	}
+	for i, esMonth := range esCosts.Children {
+		explorerMonthCostPtr := explorerCosts.ResultsByTime[i].Total["UnblendedCost"].Amount
```
thibautcornolti: API does not guarantee any month order, and probably as well as AWS Explorer.

thibautcornolti (changes requested):

server/taskCheckCost.go
zanchi-r: From my tests they are in a good order. This is only a task to help us do some debug on the trackit account so it's not a big deal if the code is not perfect. Should I still spend time to do the modifications?

thibautcornolti: Up to you, just to let you know it might be random.

thibautcornolti (approved):

simonmeyerrr (approved):

PR #199, opened 2018-12-17
[FIX] max result elastisearch

simonmeyerrr:


thibautcornolti (commented):
Probably not the right way, we should not keep two different versions. Let's keep the code intact and downgrade the staging ES to `5.5`, like our production.

simonmeyerrr:
Yes, that's why I finally put it in review only :/

PR #207, opened 2019-01-02
Anomalies warning instead of log when data not available yet

zanchi-r:
This PR updates the anomalies task to log a warning instead of an error when the data is not available yet


zanchi-r:
Already fixed but not deployed, closing this PR

PR #254, opened 2019-11-04
Fix Request API Issue

jgchoppe:
Verification of a variable which where in some different file so that doesn't crash if the variable is null.


giubil (approved):
LGTM

PR #274, opened 2020-07-01
fix/add_lambda_region

alexandre-snr:
Added region field in lambda usage reports.


aws/usageReports/lambda/mappings.go
```diff
@@ -69,6 +69,9 @@ const TemplateLineItem = `
 						"memory": {
 							"type": "integer"
 						},
+						"region": {
```
simonmeyerrr, line 81: Please, update mappings version from `1` to `2`

simonmeyerrr (changes requested):

simonmeyerrr (approved):

PR #287, opened 2020-07-23
Added tagging suggestions route

alexandre-snr:
This Pr adds the route to get 3 suggestions of values for a tag key passed in query's arguments.

Suggestions are based on the 3 most used values for the key

### DIFF
--- a/tagging/routes/routeGetTaggingSuggestions.go
+++ b/tagging/routes/routeGetTaggingSuggestions.go
@@ -27,30 +27,30 @@ import (
 	"github.com/trackit/trackit/users"
 )
 
-const maxSuggestionsCount = 3
+const maxSuggestionsCount = 10
 
 func routeGetTaggingSuggestions(r *http.Request, a routes.Arguments) (int, interface{}) {
 	u := a[users.AuthenticatedUser].(users.User)
-	tagName := a[suggestionsQueryArgs[0]].(string)
+	tagKey := a[suggestionsQueryArgs[0]].(string)
 
-	suggestions, err := getSuggestions(r.Context(), u.Id, tagName)
+	suggestions, err := getSuggestions(r.Context(), u.Id, tagKey)
 	if err != nil {
 		return http.StatusInternalServerError, nil
 	}
 
 	return http.StatusOK, map[string]interface{}{
-		"tagName":     tagName,
+		"tagKey":      tagKey,
 		"suggestions": suggestions,
 	}
 }
 
-func getSuggestions(ctx context.Context, userId int, tagName string) ([]string, error) {
+func getSuggestions(ctx context.Context, userId int, tagKey string) ([]string, error) {
 	client := es.Client
 
 	res, err := client.Search().Index(es.IndexNameForUserId(userId, tagging.IndexPrefixTaggingReport)).Size(0).
 		Aggregation("byDate", elastic.NewTermsAggregation().Field("reportDate").Order("_term", false).Size(1).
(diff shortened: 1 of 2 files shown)


jklakosz (approved):

tagging/routes/routeGetTaggingSuggestions.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func getSuggestions(ctx context.Context, userId int, tagName string) ([]string, error) {
+	client := es.Client
+
+	res, err := client.Search().Index(es.IndexNameForUserId(userId, tagging.IndexPrefixTaggingReport)).Size(0).
+		Aggregation("byDate", elastic.NewTermsAggregation().Field("reportDate").Order("_term", false).Size(1).
+			SubAggregation("nested", elastic.NewNestedAggregation().Path("tags").
+				SubAggregation("byTagName", elastic.NewFilterAggregation().Filter(elastic.NewTermQuery("tags.key", "Name")).
```
simonmeyerrr: You probably forget to replace `"Name"` by `tagName`

tagging/routes/routes.go
```diff
@@ -85,3 +95,17 @@ func init() {
...
+		http.MethodGet: routes.H(routeGetTaggingSuggestions).With(
+			db.RequestTransaction{db.Db},
+			users.RequireAuthenticatedUser{users.ViewerAsParent},
+			routes.QueryArgs(suggestionsQueryArgs),
+			routes.Documentation{
+				Summary:     "get suggestions for a tag's value",
+				Description: "Responds with suggestions for a tag's value for a user.",
+			},
+		),
+	}.H().Register("/tagging/suggestions")
```
simonmeyerrr: Since there will probably be other types of "suggestion" in the future, it would be better to use a more explicit route name. (exemple: `tagging/suggestions/tag-value` That's one example, but I think there's a better way...)

tagging/routes/routes.go
```diff
@@ -45,14 +45,24 @@ var resourcesQueryArgs = []routes.QueryArg{
 	},
 }
 
+// suggestionsQueryArgs allows to get required queryArgs params
+var suggestionsQueryArgs = []routes.QueryArg{
+	routes.QueryArg{
+		Name:        "tagName",
```
simonmeyerrr: The url part should not be capitalized.

tagging/routes/routes.go
```diff
@@ -45,14 +45,24 @@ var resourcesQueryArgs = []routes.QueryArg{
...
+		Type:        routes.QueryArgString{},
+		Description: "Tag name for suggestions",
```
simonmeyerrr: Use the `key` keyword instead of `name` for a tag. (To change everywhere)

tagging/routes/routeGetTaggingSuggestions.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/olivere/elastic"
+
+	"github.com/trackit/trackit/es"
+	"github.com/trackit/trackit/routes"
+	"github.com/trackit/trackit/tagging"
+	"github.com/trackit/trackit/users"
+)
+
+const maxSuggestionsCount = 3
```
simonmeyerrr: You can increase the number of suggestions. For cases like the `Project` key, there can be many values.

simonmeyerrr (changes requested):

SchneiderNicolas (approved):

PR #296, opened 2020-08-11
Feat/es mapping updates

alexandre-snr:
This PR refactors some ES related code: mainly moving everything related to mappings scattered in the code base into a new package.

It also adds versioning of ES index mappings:
a new MySQL keeps track of the template's version used by every index in the ES.

ES updates are only done when the task "update-es-mappings" is run.
This task tries to discover indexes which are not tracked in the Mysql table yet. It also check if there is a new version for existing index' mappings and updates them using the following technique: [Changing maping with zero downtime](https://www.elastic.co/fr/blog/changing-mapping-with-zero-downtime).

This is for review only for now.


alexandre-snr:
This branch needs to be rebased once all other pull requests are merged.

GabrielRavier:
Updated it to be based on the latest version of stg (although I did a bunch of merges to make the merging process easier rather than a rebase, but I would hope that's fine).

PR #304, opened 2020-08-19
Feature/should popup

SchneiderNicolas:
This PR adds a route that verifies if the user has access to Tagbot (marketplace subscription, stripe subscription or free trial).

And returns whether or not to display the popup that asks the user to subscribe.


tagging/routes/routeGetShouldPopup.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	if checkUserTagbotFreeTrial(customer.Created) {
+		return 200, PopupInfoResponseBody{
+			popup: false,
+		}
+	}
+	return 200, PopupInfoResponseBody{
+		popup: true,
+	}
+}
```
alexandre-snr, line 79: `CustomerIdentifier != ""` checks are not needed. Entitlement flags are false by default if there are none.
Also could you replace `200` codes  by `http.StatusOK` please ?

alexandre-snr (changes requested):

alexandre-snr (approved):
LGTM

tagging/routes/routes.go
```diff
@@ -93,3 +93,16 @@ func init() {
...
 }
+
+func init() {
+	routes.MethodMuxer{
+		http.MethodGet: routes.H(shouldPopup).With(
+			db.RequestTransaction{db.Db},
+			users.RequireAuthenticatedUser{users.ViewerAsParent},
+			routes.Documentation{
+				Summary:     "get list of resources",
+				Description: "Responds with the list of resources based on the queryparams passed to it",
```
hug33k: Description does not reflect the purpose of this endpoint

hug33k (changes requested):

hug33k (approved):

PR #309, opened 2020-08-24
Exclude AWSDataTransfer product from /costs route

alexandre-snr:
This PR excludes `AWSDataTransfer` product in the `/costs` route.
This is because this product code is only used for taxes and the value is always 0.

Data transfer costs are actually included in other product codes: for example EC2 instances data transfer costs have a `productCode` = `AmazonEC2` and a `serviceCode` = `AWSDataTransfer`.

### DIFF
--- a/costs/es_request_constructor.go
+++ b/costs/es_request_constructor.go
@@ -265,14 +265,14 @@ func nestAggregation(allAggrSlice []paramAggrAndName) elastic.Aggregation {
 //	- If a param in the slice is not present in the detailedLineItemsFieldsName, the program will crash.
 //	- If the client is nil or malconfigured, it will crash
 //	- If the index is not an index present in the ES, it will crash
+// We are excluding AWSDataTransfer products because it's value is always zero.
+// Data transfer costs are included in other products' costs.
 func GetElasticSearchParams(accountList []string, durationBegin time.Time,
 	durationEnd time.Time, params []string, client *elastic.Client, index string) *elastic.SearchService {
 	query := elastic.NewBoolQuery()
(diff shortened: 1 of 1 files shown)


costs/es_request_constructor.go
```diff
@@ -271,7 +271,8 @@ func GetElasticSearchParams(accountList []string, durationBegin time.Time,
 	if len(accountList) > 0 {
 		query = query.Filter(createQueryAccountFilter(accountList))
 	}
-	query = query.Filter(createQueryTimeRange(durationBegin, durationEnd))
+	query = query.Filter(createQueryTimeRange(durationBegin, durationEnd),
+		elastic.NewBoolQuery().MustNot(elastic.NewTermQuery("productCode", "AWSDataTransfer")))
```
hug33k, line 277: Please add a comment above the function to explain why we have this query

hug33k (changes requested):

costs/es_request_constructor.go
```diff
@@ -271,7 +271,10 @@ func GetElasticSearchParams(accountList []string, durationBegin time.Time,
 	if len(accountList) > 0 {
 		query = query.Filter(createQueryAccountFilter(accountList))
 	}
-	query = query.Filter(createQueryTimeRange(durationBegin, durationEnd))
+	// We are excluding AWSDataTransfer products because it's value is always zero.
+	// Data transfer costs are included in other products' costs.
```
hug33k: "above the function" :)

hug33k (changes requested):

hug33k (approved):

PR #320, opened 2021-03-18
Tasks Worker

### LINKED ISSUE #319: Create the TrackIt task worker
The worker will be able to retrieve tasks from the SQS queue and achieve those ones.

Specifications:

- [x] add a worker flag (--worker=true by example), if the flag is true then run the worker instead of the API
- [x] check if ES is available
- [x] listens to the SQS queue to retrieve tasks
- [x] run the task
- [x] handle task success (send acknowledge to the queue) or error (send changes on visibility timeout to ~15min to re-run the job later)
- [x] send logs on cloudwatch groups logs (separated by task by account: `${task_name}/${account_id}`

jgchoppe:
The worker is going to handle tasks from the AWS SQS queue


server/taskWorker.go
```diff
@@ -0,0 +1,298 @@
...
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"github.com/trackit/jsonlog"
+	"github.com/trackit/trackit/es"
```
simonmeyerrr: misplaced import

server/server.go
```diff
@@ -88,7 +88,9 @@ func main() {
 	logger.Info("Started.", struct {
 		BackendId string `json:"backendId"`
 	}{backendId})
-	if task, ok := tasks[config.Task]; ok {
+	if config.Worker {
```
simonmeyerrr, line 91: Instead of adding a new parameter, you could just create a task with the name `worker` as it is the same kind of handler. So it could be used like `./server --task worker`

server/taskWorker.go
```diff
@@ -0,0 +1,298 @@
...
+			continue
+		}
+
+		logger.Info("Received message, checking ES health and executing task.", map[string]interface{}{
+			"message": message,
+		})
+
+		if res, err := es.Client.ClusterHealth().Timeout(strconv.Itoa(esHealthcheckTimeoutInMinutes) + "m").Do(ctx); err != nil || res.TimedOut {
+			logger.Error("ES is not reachable.", map[string]interface{}{
+				"timeout": strconv.Itoa(esHealthcheckTimeoutInMinutes) + "m",
```
simonmeyerrr: `strconv.Itoa(esHealthcheckTimeoutInMinutes) + "m"` could be in a var as it is used multiple times

server/taskWorker.go
```diff
@@ -0,0 +1,298 @@
...
+			})
+			logsBuffer.Reset()
+			_ = resetMessageTimeout(ctx, sqsq, queueUrl, receiptHandle)
+			continue
+		}
+
+		ctx = context.WithValue(ctx, "taskParameters", message.Parameters)
+
+		if task, ok := tasks[message.TaskName]; ok {
+			err = task(ctx)
```
simonmeyerrr: It could be nice to catch `panic` to be able to manage errors.

server/taskWorker.go
```diff
@@ -0,0 +1,298 @@
...
+		return messageData, nil, err
+	}
+	return messageData, msgResult.Messages[0].ReceiptHandle, nil
+}
+
+func flushCloudwatchLogEvents(ctx context.Context, cwl *cloudwatchlogs.CloudWatchLogs, message MessageData, logsBuffer *bytes.Buffer) error {
+	logger := jsonlog.LoggerFromContextOrDefault(ctx)
+
+	if config.Environment != "prod" && config.Environment != "stg" {
+		logsBuffer.Reset()
```
simonmeyerrr, line 169: You can use `defer` instead of having this line before each return

server/taskWorker.go
```diff
@@ -0,0 +1,298 @@
...
+		})
+		logsBuffer.Reset()
+		return err
+	}
+
+	logInputs := decodeLogBuffer(*logsBuffer)
+
+	var logEvents []*cloudwatchlogs.InputLogEvent
+	for _, logInput := range logInputs {
+		fmt.Println(logInput.Message)
```
simonmeyerrr: Forgotten debug print

simonmeyerrr (changes requested):
Some changes to do.

What happens if a task takes more than 30 minutes to complete? From what I understand, the SQS message is not deleted because the task is in progress and will be visible in the queue, which can restart the same task a second time. Or it is managed but I don't understand it?

simonmeyerrr (dismissed):

simonmeyerrr (approved):

PR #332, opened 2021-07-22
Added some general documentation

GabrielRavier:
I've cleaned up some of the text from the old documentation branch (changed it to be up-to-date in some places).


documentation/models.md
```diff
@@ -0,0 +1,31 @@
+# Models
+
+In order to interact with the SQL database, why use `xo` which is a command-line tool used to generate Go code based on a database schema.
```
cruffinoni: ```why use `xo` which is a command-line tool``` is "**why**" intended ?

cruffinoni (changes requested):

documentation/models.md
GabrielRavier: Seems like it probably wasn't, should be fixed now.

J3suis (approved):
LGTM

cruffinoni (approved):

PR #333, opened 2021-07-27
Fix a bunch of context leaks

GabrielRavier:
Context leaks are very frequent because of one copy-pasted code template that only executes the context cancel function when there is no error (meaning there is a leak, however temporary because of the WithTimeout usage, whenever an error occurs). There is also one other leak of this kind which would always occur, in `aws/s3/readBills.go`, which also has been fixed.


J3suis (approved):
LGTM

cruffinoni (approved):

EthanHusser (approved):

simonmeyerrr (approved):

PR #335, opened 2021-07-28
Oddly written (including unreachable code) function in aws/routes/routeAwsSubaccountPath.go

GabrielRavier:
There's some unreachable code in patchAwsSubAccountRequestBody which just answers with an 200 OK response code and nothing else (whereas it's supposed to send back the AWS account information). I've
refactored the code around it too as the if-else usage in general was somewhat dubious too.


cruffinoni (approved):
LGTM but we will wait for @simonmeyerrr to review this one before merging since it's a different coding style.

simonmeyerrr (approved):

PR #336, opened 2021-07-29
Fix some bad format strings

GabrielRavier:
Several fmt.Sprintln calls would use "%s" erroneously (fmt.Sprintln doesn't interpret format strings), which would result in an output containing a random %s in the middle after the ": ".

There was also one instance in which an fmt.Sprintf call used %u, which is not a valid verb in Go (%v should be used instead)


cruffinoni (approved):
LGTM

simonmeyerrr (approved):

PR #337, opened 2021-07-29
Run gofmt on all files

GabrielRavier:
It seems like there are a lot of formatting problems (at least according to gofmt, which is the official Go formatter they recommend everybody use), so I've gone and fixed them with gofmt


cruffinoni (approved):
I'm pretty sure we'll have problems when this RP is merged as it affects a lot of files. Make sure you keep an eye out for conflicts whenever another RP is merged.

simonmeyerrr (approved):

PR #338, opened 2021-07-29
Feat/update most used tags

J3suis:
- Add a ways to see all mostusedtags history by user with routes /tagging/mostusedtags-history
- Link most used tags to /tagging/compliance response

### DIFF
--- a/tagging/routes/routeGetMostUsedTags.go
+++ b/tagging/routes/routeGetMostUsedTags.go
@@ -17,6 +17,7 @@ package routes
 import (
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"net/http"
 
 	"github.com/trackit/jsonlog"
@@ -63,13 +64,11 @@ func routeGetMostUsedTagsHistory(r *http.Request, a routes.Arguments) (int, inte
 	dbRes, err := models.MostUsedTagsHistoryByUser(tx, u.Id)
 	if err != nil {
 		logger.Error("Could not fetch most used tags.", err.Error())
-		return http.StatusInternalServerError, nil
+		return http.StatusInternalServerError, err.Error()
 	}
 
 	if dbRes == nil {
-		return http.StatusInternalServerError, map[string]interface{}{
-			"error": "No history reports available.",
-		}
+		return http.StatusInternalServerError, errors.New("no history reports available")
 	}
 	history := make([]map[string]interface{}, 0)
 	for _, mut := range dbRes {
--- a/tagging/routes/routeGetMostUsedTags.go
+++ b/tagging/routes/routeGetMostUsedTags.go
@@ -15,6 +15,7 @@
 package routes
 
 import (
+	"database/sql"
 	"encoding/json"
 	"net/http"
 
@@ -29,8 +30,8 @@ import (
 func routeGetMostUsedTags(r *http.Request, a routes.Arguments) (int, interface{}) {
 	logger := jsonlog.LoggerFromContextOrDefault(r.Context())
 	u := a[users.AuthenticatedUser].(users.User)
-
-	dbRes, err := models.MostUsedTagsInUseByUser(db.Db, u.Id)
+	tx := a[db.Transaction].(*sql.Tx)
+	dbRes, err := models.MostUsedTagsInUseByUser(tx, u.Id)
 	if err != nil {
 		logger.Error("Could not fetch most used tags.", err.Error())
 		return http.StatusInternalServerError, nil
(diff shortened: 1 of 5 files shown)


tagging/routes/routeGetMostUsedTags.go
```diff
@@ -54,3 +55,37 @@ func routeGetMostUsedTags(r *http.Request, a routes.Arguments) (int, interface{}
...
 }
+
+func routeGetMostUsedTagsHistory(r *http.Request, a routes.Arguments) (int, interface{}) {
+	logger := jsonlog.LoggerFromContextOrDefault(r.Context())
+	u := a[users.AuthenticatedUser].(users.User)
+	tx := a[db.Transaction].(*sql.Tx)
+	dbRes, err := models.MostUsedTagsHistoryByUser(tx, u.Id)
+	if err != nil {
+		logger.Error("Could not fetch most used tags.", err.Error())
+		return http.StatusInternalServerError, nil
```
GabrielRavier: From what I can see in other code, you can also return the details of the error as the second return value is an `interface{}` and something like `error` will be handled perfectly well, as can be seen from code in `users/create.go`, for example.

tagging/routes/routeGetMostUsedTags.go
```diff
@@ -54,3 +55,37 @@ func routeGetMostUsedTags(r *http.Request, a routes.Arguments) (int, interface{}
...
+	}
+
+	if dbRes == nil {
+		return http.StatusInternalServerError, map[string]interface{}{
```
GabrielRavier: Same here, you don't have to return a `map[string]interface{}`, you can just go and return something like `errors.New("No history reports available")` (unless the `map[string]interface{}` looks much better or something like that)

tagging/routes/routes.go
```diff
@@ -56,6 +56,17 @@ func init() {
...
 	}.H().Register("/tagging/mostusedtags")
 
+	routes.MethodMuxer{
+		http.MethodGet: routes.H(routeGetMostUsedTagsHistory).With(
+			db.RequestTransaction{db.Db},
+			users.RequireAuthenticatedUser{users.ViewerAsParent},
+			routes.Documentation{
+				Summary:     "get most used tags",
+				Description: "Responds with most used tags for a user.",
+			},
```
GabrielRavier, line 66: This documentation seems inaccurate, considering it's a copy-paste of the one from `/tagging/mostusedtags`

GabrielRavier (changes requested):

GabrielRavier (approved):
LGTM now

tagging/routes/routeGetMostUsedTags.go
```diff
@@ -54,3 +56,35 @@ func routeGetMostUsedTags(r *http.Request, a routes.Arguments) (int, interface{}
...
+	u := a[users.AuthenticatedUser].(users.User)
+	tx := a[db.Transaction].(*sql.Tx)
+	dbRes, err := models.MostUsedTagsHistoryByUser(tx, u.Id)
+	if err != nil {
+		logger.Error("Could not fetch most used tags.", err.Error())
+		return http.StatusInternalServerError, err.Error()
+	}
+
+	if dbRes == nil {
+		return http.StatusInternalServerError, errors.New("no history reports available")
```
simonmeyerrr: Should be a `404` error

tagging/routes/routeGetMostUsedTags.go
```diff
@@ -54,3 +56,35 @@ func routeGetMostUsedTags(r *http.Request, a routes.Arguments) (int, interface{}
...
+	}
+	history := make([]map[string]interface{}, 0)
+	for _, mut := range dbRes {
+		tagsList := []string{}
+		err = json.Unmarshal([]byte(mut.Tags), &tagsList)
+		if err != nil {
+			logger.Error("Could not unmarshal most used tags.", err.Error())
+			return http.StatusInternalServerError, err
```
simonmeyerrr: Use `continue` so a little error does not block the user to get valid history data.

tagging/routes/routeGetMostUsedTags.go
```diff
@@ -54,3 +56,35 @@ func routeGetMostUsedTags(r *http.Request, a routes.Arguments) (int, interface{}
...
+		}
+		if mut == nil || len(tagsList) == 0 {
```
simonmeyerrr: It's to late to check if `mut` is nil as you already use it earlier

simonmeyerrr (changes requested):

simonmeyerrr (approved):

PR #339, opened 2021-07-30
Fix a bunch of shadowing problems

GabrielRavier:
A bunch of these changes aren't straight up bug fixes (though most of them are), but they keep the shadowing to a minimum while refactoring the code to better correspond to the recommended guidelines on writing Go code. Otherwise, most of the bug fixes are about `err` getting shadowed before further usage, which can be pretty bad when it results in no error being returned in a situation where one occured.

Note that while `:=` will usually only redeclare (and thus not truly shadow) an existing identifier, this does not apply when the identifier is not from the same scope. `if` statements do in fact introduce a new scope even in the preceding statement (i.e. `stmt` in `if stmt; cond {}` in not in the same scope as the code containing the `if`), which is the cause of many of the bugs here.

### DIFF
--- a/anomaliesDetection/products.go
+++ b/anomaliesDetection/products.go
@@ -234,7 +234,7 @@ func productGetAnomaliesData(ctx context.Context, params AnomalyEsQueryParams) (
 		return nil, err
 	}
 	var typedDocument esProductTypedResult
-	if err := json.Unmarshal(*sr.Aggregations["products"], &typedDocument.Products); err != nil {
+	if err = json.Unmarshal(*sr.Aggregations["products"], &typedDocument.Products); err != nil {
 		logger.Error("Failed to parse elasticsearch document.", err.Error())
 		return nil, err
 	}
@@ -260,8 +260,5 @@ func productGetAnomaliesData(ctx context.Context, params AnomalyEsQueryParams) (
 		totalAnalyzedCosts = append(totalAnalyzedCosts, aCosts...)
 	}
 	totalAnalyzedCosts = productClearDisturbances(totalAnalyzedCosts, totalCostsByDay, highestSpendersByDay)
-	if err != nil {
-		return nil, err
-	}
 	return totalAnalyzedCosts, nil
 }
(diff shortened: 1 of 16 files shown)


anomaliesDetection/products.go
```diff
@@ -260,8 +260,5 @@ func productGetAnomaliesData(ctx context.Context, params AnomalyEsQueryParams) (
 		totalAnalyzedCosts = append(totalAnalyzedCosts, aCosts...)
 	}
 	totalAnalyzedCosts = productClearDisturbances(totalAnalyzedCosts, totalCostsByDay, highestSpendersByDay)
-	if err != nil {
```
cruffinoni, line 265: Why has this part of the code been removed?

es/awsClientSigner.go
```diff
@@ -50,16 +50,16 @@ func checkParametersError(endpointMetaData []string, creds *credentials.Credenti
...
-		} else {
-			return nil, err
-		}
-	} else {
+	cofs, err := NewSignedElasticClientOptions(endpoint, creds)
+	if err != nil {
 		return nil, err
 	}
+	cof := configEach(cofs...)
+	ec, err := elastic.NewClient(elastic.SetURL(endpoint), cof)
```
cruffinoni, line 58: This function has no real problem with variable shadowing and you also do it on this modification

anomaliesDetection/products.go
GabrielRavier: It could stay, but it's not doing anything: `err` will always be `nil` when it is reached. I suppose this might be more appropriate for another PR, should you consider it very important to put it somewhere else like that.

es/awsClientSigner.go
GabrielRavier: I also wanted to remove as many occurrences of shadowing, however innocent they might actually be, as possible (so long as it doesn't require some massive code change or something), so that finding future problems with usage of shadowing would be simpler, but if you want I can revert this change.

anomaliesDetection/products.go
cruffinoni: You right 👍

es/awsClientSigner.go
cruffinoni: Both versions are correct but the logic is reversed. What do you think @simonmeyerrr ?

GabrielRavier: (also, I cannot seem to find any formally written traces of this but I believe using `if err == nil` is considered an anti-pattern in general by the Go devs and they recommend inversing it like I did here where possible)

GabrielRavier: ([Found it recommended on the official repo](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow))

simonmeyerrr (approved):

PR #340, opened 2021-08-02
Fix comparisons with boolean constants

GabrielRavier:
Not exactly major bugs, but I saw it mentioned as recommended style directly from the Go authors so I went and fixed occurrences of it


cruffinoni (approved):
LGTM

simonmeyerrr (approved):

PR #342, opened 2021-08-04
General cleanup and optimization of minor issues

GabrielRavier:
Numerous issues are fixed:
- Unmerged adjacent declarations and assignments (`anomaliesDetection`, `aws`)
- Unused values of certain variables (`costs`, `s3`, `users`)
- Redundant returns (`reports`)
- Bad usage of loops instead of just doing `append(a, b...)` (much simpler and very likely faster) (`cache`, `reports`, `routes`)
- Using `errors.New(fmt.Sprintf` instead of just `fmt.Errorf` (`aws`, `reports`, `users`)
- Using `fmt.Sprintf` when unnecessary (`aws`, `tagging`, `users`)
- Mispelled or missing type or func names in comments (`aws`, `costs`, `db`, `es`, `onDemandToRI`, `routes`, `tagging`, `usageReports`, `users`)
- Code that was way too verbose (`tagging.checkUserTagbotFreeTrial`)
- Using `time.Now().Sub` instead of `time.Since` (`routes`, `tagging`, `unusedAccounts`, `users`)
- Using `t.Add(time.Now())` instead of `time.Until(t)` (`unusedAccounts`)
- Unnecessary checks for `nil`, primarily w.r.t. guarding uses of `len` (`usageReports`, `users`, `util`)
- Useless use of `make(map[T]T, 0)` (often corrected to a more accurate size) (`aws`, `models`, `usageReports`)
- Unnecessary usage of `_` where no identifier is needed (`aws`, `costs`, `plugins`, `users`, `routes`)
- Useless usage of `break` (`errors`)
- Bad usage of type assertions that could be optimized (`errors`)
- Useless code that uses up time with no used results (`aws`)
- Fix some empty branches (`anomaliesDetection`, `server`)
- Usage of `+= 1` instead of `++` (`aws`, `costs`, `onDemandToRI`, `plugins`)
- Uncommented usage of blank imports that should probably be at least documented a bit (`db`, `plugins`)

### DIFF
--- a/aws/routes/routeAwsStatus.go
+++ b/aws/routes/routeAwsStatus.go
@@ -30,7 +30,7 @@ func getAwsAccountsStatus(r *http.Request, a routes.Arguments) (int, interface{}
 		l.Error("failed to get AWS accounts' bill repositories", err.Error())
 		return 500, errors.New("failed to retrieve bill repositories")
 	}
-	// Code unneeded for now
+	// Code unneeded for now considering all it does is create an array and fill it with all the id data from the AWS stuff before proceeding to do nothing with it
 	/*
 	billRepositoriesIds := make([]int, 0)
 	for _, awsAccount := range awsAccountsWithBillRepositories {
--- a/aws/routes/routeAwsStatus.go
+++ b/aws/routes/routeAwsStatus.go
@@ -30,12 +30,15 @@ func getAwsAccountsStatus(r *http.Request, a routes.Arguments) (int, interface{}
 		l.Error("failed to get AWS accounts' bill repositories", err.Error())
 		return 500, errors.New("failed to retrieve bill repositories")
 	}
+	// Code unneeded for now
+	/*
 	billRepositoriesIds := make([]int, 0)
 	for _, awsAccount := range awsAccountsWithBillRepositories {
 		for _, billRepository := range awsAccount.BillRepositories {
 			billRepositoriesIds = append(billRepositoriesIds, billRepository.Id)
 		}
 	}
+	*/
 	result := s3.WrapAwsAccountsWithBillRepositoriesWithPendingWithStatus(awsAccountsWithBillRepositories, tx)
 	return 200, result
 }
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -27,15 +27,15 @@ import (
 func GetErrorMessage(ctx context.Context, err error) error {
 	logger := jsonlog.LoggerFromContextOrDefault(ctx)
 	var formattedErr error
-	switch err.(type) {
+	switch err := err.(type) {
 	case *elastic.Error:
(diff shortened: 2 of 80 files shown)


cruffinoni:
Adding the label "high priority". It fixes many bugs.

aws/routes/routeAwsStatus.go
```diff
@@ -30,12 +30,15 @@ func getAwsAccountsStatus(r *http.Request, a routes.Arguments) (int, interface{}
 		l.Error("failed to get AWS accounts' bill repositories", err.Error())
 		return 500, errors.New("failed to retrieve bill repositories")
 	}
+	// Code unneeded for now
```
cruffinoni: Can you explain why this code is no more needed, please?

errors/errors.go
```diff
@@ -27,15 +27,15 @@ import (
 func GetErrorMessage(ctx context.Context, err error) error {
 	logger := jsonlog.LoggerFromContextOrDefault(ctx)
 	var formattedErr error
-	switch err.(type) {
+	switch err := err.(type) {
```
cruffinoni, line 30: This modification is debatable

s3/costs/prepare_response.go
```diff
@@ -108,7 +108,7 @@ func parseBuckets(buckets BucketsInfo, parsedDocument bucket, resultType string)
 		if isValidBucket(bucketName) {
 			bucketInfo := getBucketInfoByName(buckets, bucketName)
```
cruffinoni: `bucketInfo` could be created directly after the condition below, couldn't it?

aws/routes/routeAwsStatus.go
GabrielRavier: Made the comment clearer.

Btw do tell if you think the code should simply be removed, I kept it (though commented) in the case it was there to be used in some later functionality.

errors/errors.go
GabrielRavier: How, exactly ? It seems to me like it simplifies the code nicely (it's probably also a minor optimization now that I think about it) compared to manual type assertions for every case... If you want, I can change the variable name if you find that it makes things unclear

s3/costs/prepare_response.go
GabrielRavier: Done.

errors/errors.go
cruffinoni: After a second look, I do agree with your changes. My bad.

aws/routes/routeAwsStatus.go
cruffinoni: It is better to explain directly in the code why this is not necessary at the moment. We will keep it, however. You never know.

cruffinoni (approved):
Great, looks good. We will wait for @simonmeyerrr or someone else to review it, as it's a huge PR.

simonmeyerrr (approved):

PR #343, opened 2021-08-18
Added discount codes for TagBot

GabrielRavier:
Discount codes have been added:
- A new `tagbot_discount_code` table has been added to the SQL database, and properly integrated with `tagbot_user`
- A `discount_code` field has been added to the user creation request body, which (if non-empty) is checked for being a valid existing discount code and changes the free trial length from 14 days to 2 months

There are also some major changes in the xo generated code as the latest version has some changes in the generator (including a few externally incompatible ones I fixed) that seem to make it impossible to combine code from it with code from older versions. I also went and fixed the bad usage of a few `created` fields while I was at it, except for `forgotten_password` where it seems to be required.


cruffinoni:
This is a good PR but you need to split the feature from the changes concerning the database.

GabrielRavier:
What changes do you mean, exactly ? Do you mean:
- the directly-related additions to the database to enable discount codes to be stored in there that are fundamental to the feature ? (really cannot be removed without making the feature completely disfunctional)
- the changes I had to make to all the generated code so as to be able to use `xo` with it ? (could be removed but it would be a mess and I'd have to either try to make the two `xo` worlds work without knowledge of each other, which would likely require tweaking the generated code), or I'd have to manually recreate what I think `xo` would generate for the new discount codes changes (i.e. guessing what the generated code would have been with the old version of `xo`)

simonmeyerrr (approved):

PR #348, opened 2021-08-25
Make db/schema.sql and db/migrations completely consistent

GabrielRavier:
Before this, it seems like they were fully functionally equivalent, but making it so in a way that can be checked in a single command is a lot more useful and makes checking/testing for equivalence a lot easier.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -242,7 +242,7 @@ ALTER TABLE aws_account_update_job ADD ec2Error VARCHAR(255) NOT NULL DEFAULT ""
 
 ALTER TABLE aws_bill_repository ADD grace_update DATETIME NOT NULL DEFAULT "1970-01-01 00:00:00";
 
---   Copyright 2017 MSolution.IO
+--   Copyright 2018 MSolution.IO
 --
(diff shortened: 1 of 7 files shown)


db/schema.sql
```diff
@@ -238,7 +242,7 @@ ALTER TABLE aws_account_update_job ADD ec2Error VARCHAR(255) NOT NULL DEFAULT ""
 
 ALTER TABLE aws_bill_repository ADD grace_update DATETIME NOT NULL DEFAULT "1970-01-01 00:00:00";
 
---   Copyright 2018 MSolution.IO
+--   Copyright 2017 MSolution.IO
```
cruffinoni: Why has this been changed?

cruffinoni (commented):
That's a very good PR

db/schema.sql
GabrielRavier: It was changed because that's what was in the file, but it seems as though the file was inaccurate and the copyright date should be changed to 2018, so I've gone and done that

simonmeyerrr (approved):

PR #349, opened 2021-08-25
Cleaned up a *lot* of places where errors are ignored or otherwise badly handled

GabrielRavier:
Note: You might want to see whether it might be better in some places to throw the error back up the call stack to end up being reported elsewhere, I might have possibly been overly careful in some places as I wanted to avoid the chances of false positives as much as possible (i.e. there's a bunch of places where I just log the error and do nothing else with it).

I also corrected a few task name typos in server while I was at it since they were right next to the errors, but do tell me if that should go into another PR

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -1,4 +1,4 @@
-//   Copyright 2017 MSolution.IO
+//   Copyright 2021 MSolution.IO
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
 //   you may not use this file except in compliance with the License.
--- a/server/utils.go
+++ b/server/utils.go
@@ -1,0 +1,43 @@
+//   Copyright 2017 MSolution.IO
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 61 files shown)


server/utils.go
```diff
@@ -0,0 +1,43 @@
+//   Copyright 2017 MSolution.IO
```
cruffinoni: Shouldn't you put 2021?

users/password.go
```diff
@@ -78,18 +78,30 @@ func init() {
...
+	}
 	if err == nil {
-		now := time.Now()
-		expire := now.Add(time.Hour * time.Duration(-1*int64(nbHoursValidityForgottenToken)))
-		models.DeleteExpiredForgottenPassword(transaction, expire)
 		defer func() {
 			rec := recover()
-			if rec != nil {
-				transaction.Rollback()
+			if rec != nil || err != nil {
```
cruffinoni: `err` will be always null in this condition

cruffinoni (changes requested):
I'm waiting for @simonmeyerrr or someone else to do a review before merging.

server/utils.go
GabrielRavier: Oops, that was a copy-paste error. Corrected it !

users/password.go
GabrielRavier: It's in a deferred function so it can be non-nil after the call to `models.DeleteExpiredForgottenPassword`

simonmeyerrr (approved):

PR #350, opened 2021-08-26
Fixed loads of spelling errors

GabrielRavier:
A majority of these are in comments but a lot of them are in error messages and often in stuff that'll be returned to the web client and visible to its user

### DIFF
--- a/aws/usageReports/riEc2/utils.go
+++ b/aws/usageReports/riEc2/utils.go
@@ -62,7 +62,7 @@ type (
 		Tags []utils.Tag `json:"tags"`
 	}
 
-	//RecurringCharges contains recurring charges informations of a reservation
+	// RecurringCharges contains recurring charges information for a reservation
 	RecurringCharges struct {
 		Amount    float64
 		Frequency string
--- a/aws/usageReports/riRdS/utils.go
+++ b/aws/usageReports/riRdS/utils.go
@@ -60,7 +60,7 @@ type (
 		Tags []utils.Tag `json:"tags"`
 	}
 
-	//RecurringCharges contains recurring charges informations of a reservation
+	// RecurringCharges contains recurring charges information for a reservation
 	RecurringCharges struct {
 		Amount    float64
 		Frequency string
--- a/routes/constError.go
+++ b/routes/constError.go
@@ -14,7 +14,7 @@
 
 package routes
 
-// constError is a string which fullfils the error interface.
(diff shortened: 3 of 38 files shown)


aws/usageReports/riEc2/utils.go
```diff
@@ -62,7 +62,7 @@ type (
 		Tags []utils.Tag `json:"tags"`
 	}
 
-	//RecurringCharges contains recurring charges informations of a reservation
+	// RecurringCharges contains recurring charges information for a reservation
```
cruffinoni, line 65: I'm not sure about this one.

aws/usageReports/riRdS/utils.go
```diff
@@ -60,7 +60,7 @@ type (
 		Tags []utils.Tag `json:"tags"`
 	}
 
-	//RecurringCharges contains recurring charges informations of a reservation
+	// RecurringCharges contains recurring charges information for a reservation
```
cruffinoni, line 63: Same as above, I'm not sure about this one.

routes/constError.go
```diff
@@ -14,7 +14,7 @@
 
 package routes
 
-// constError is a string which fullfils the error interface.
+// constError is a string which fulfills the error interface.
```
cruffinoni, line 17: `fills` can also be used here. What do you think?

cruffinoni (changes requested):

cruffinoni (commented):
I'm not asking for changes but I am questioning some of them. There is no approval or disapproval here, it's my fault!

routes/constError.go
GabrielRavier: I think that "fulfilling an interface" makes more sense than filling it

aws/usageReports/riRdS/utils.go
GabrielRavier: I'll return to this one once we've handled the one above

aws/usageReports/riEc2/utils.go
GabrielRavier: What do you mean, exactly ? I'm pretty sure information can't be used in plural (except when talking about court filings, but that's not exactly the case here). Though I guess it might be "the recurring charges information of a reservation" rather than "recurring charges information for a reservation" but it seems more likely to me to be the latter

simonmeyerrr (approved):

simonmeyerrr (approved):

PR #368, opened 2023-11-21
Fix: add tagbot_onboarding and tagbot_onboarding_started columns and update status at the end of the job

jgchoppe:


alexandre-snr (changes requested):
Could you revert all the changes that were made to the xo files that should not have been modified?

alexandre-snr (approved):
LGTM, I think there is another PR on another repo as well so not merging yet?

jgchoppe:
> LGTM, I think there is another PR on another repo as well so not merging yet?

I think we can already merge as those changes won't change the current workload.
The other PR is here: https://github.com/trackit/trackit-k8s-dispatcher/pull/3

